PR #13, opened 2015-09-01
single line breaks in listings package

mungerd:
This replaces double line breaks with single ones in the member variable `plain_listing` in classes from `Packages/listings.py`.

Explanation: When iterating over the lines of a file, the newline characters are not stripped from the iterated variable.  So, appending a newline doubles the line breaks.

### DIFF
--- a/plasTeX/Packages/listings.py
+++ b/plasTeX/Packages/listings.py
@@ -94,7 +94,11 @@ def _format(self, file):
             line = re.sub('/\*@[^@]*@\*/', '', line)
 
             # Add the just-read line to the listing.
-            self.plain_listing += line
+            if hasattr(file, 'read'):
+                self.plain_listing += line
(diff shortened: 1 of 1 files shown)


tiarno:
hi, thanks for catching that bug! 
Your fix works for the case when you're reading from a file, but the function is also called with strings that have no newlines. What about adding a test like this:

    if hasattr(file, 'read'):
        self.plain_listing += line
    else:
        self.plain_listing += '\n' + line

It could be fixed in a few different ways--Kevin what do you think?

kesmit13:
That seems reasonable to me.

tiarno:
@mungerd, what do you think? Can you make that change to your pull request?

mungerd:
Ok, I've made the change.  I've tested it with input from files.

PR #17, opened 2015-09-10
support nested maths in MathShift

mungerd:
This is useful when using commands that turn off math mode, such as `\cases` or AMS's `\text`.

The old implementation would consider the current MathShift command as a closing one whenever there already was an open math environment.  This led to a bug in
`$$ x = \text{$x$} $$`
as well as in
`$$ x = \cases{x & if $x=x$} $$`

The new implementation regards the current MathShift command as a closing command *only if it matches the last opening command*.  This should not break the parsing of correct LaTeX code.

### DIFF
--- a/plasTeX/Base/TeX/Primitives.py
+++ b/plasTeX/Base/TeX/Primitives.py
@@ -76,32 +76,22 @@ class MathShift(Command):
 
         """
         inEnv = type(self).inEnv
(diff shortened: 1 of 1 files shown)


mungerd:
Sorry, this breaks something I don't yet understand.  I'm going to cancel this.

PR #18, opened 2015-09-10
support nested maths in MathShift

mungerd:
This is useful when using commands that turn off math mode, such as the old `\cases` macro or AMS's `\text`.

The old implementation would consider the current MathShift command as a closing one whenever there already was an open math environment.  This led to a bug in
`$$ x = \text{$x$} $$`
as well as in
`$$ x = \cases{x & if $x=x$} $$`

The new implementation regards the current MathShift command as a closing command *only if it matches the last opening command*.  This should not break the parsing of correct LaTeX code.

Successfully passed a regression test on a project of 477 LaTeX files.

### DIFF
--- a/plasTeX/Base/TeX/Primitives.py
+++ b/plasTeX/Base/TeX/Primitives.py
@@ -76,32 +76,23 @@ class MathShift(Command):
 
         """
         inEnv = type(self).inEnv
-        math = self.ownerDocument.createElement('math')
-        displaymath = self.ownerDocument.createElement('displaymath')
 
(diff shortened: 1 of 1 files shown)


tiarno:
Thanks for the pull request, and especially thanks for the regression testing. We just can't get enough of that stuff. Merging it in, with Kevin's blessing.

tiarno:
ah, now it is.

mungerd:
You're welcome.  PlasTeX was very useful to me lately.

Issue #31, opened 2016-07-02
Renderer configuration


PatrickMassot:
Is there any way for a renderer to define config options (command line or config file)?

If no, would you be interested in a pull request where the function ``main`` in ``plastex`` asks all renderers for config options they'd like to declare, each in its own section of the ``ConfigManager`` object?

kesmit13:
That's something we had planned a long time ago, but just never got around to.  I'd be interested in seeing what you had in mind.

PR #32, opened 2016-07-03
Allow renderers to define config options

PatrickMassot:
This PR tackles issue #31 (which is maybe what is referred to in the TODO file: "Renderer specific config files"). 

The goal is to allow renderers to define options (either command line or config files). 

### Interface

Each renderer directory can contain a ``Config.py`` file which defines a variable ``config`` which is a ``ConfigManager`` instance. For instance, one can add a file ``plasTeX/Renderer/XHTML/Config.py`` containing:
```python
from plasTeX.ConfigManager import *

config = ConfigManager()

section = config.add_section('xhtml')

config.add_category('xhtml', 'XHTML renderer Options')

section['test'] = StringOption(
    """ Test option """,
    options='--test',
    category='xhtml',
    default='Yop',
)
```

When the main ``plastex`` script is invoked, it will add the ``test`` option to the recognized options, and its value will be available when rendering.

### Implementation

Everything happens inside the main script, we simply try to import a ``Config`` submodule from each renderer and merge it using the ability of ``ConfigManager`` objects to be merged (using ``__iadd__``).

### Backward compatibility

This PR is fully backward compatible, renderers are free not to implement a ``Config`` submodule.

### DIFF
--- a/plasTeX/plastex
+++ b/plasTeX/plastex
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import os, sys, codecs, string, glob
+import importlib
 import plasTeX
 from plasTeX.TeX import TeX
 from plasTeX.Config import config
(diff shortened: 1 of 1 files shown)


PatrickMassot:
Thanks for merging. We need to document this though. I'm willing to try but I'd like to know whether you have an opinion about where to put this documentation. It could be in chapter 2, 5 or 6 (or mentioned in several places).

kesmit13:
I'd probably put it in the example in chapter 5 to introduce it, but add the reference documentation to chapter 6.

Issue #28, opened 2016-06-27
newcounter "the*" bug


PatrickMassot:
It seems that the `newcounter` class is broken when the name of the counter starts with "the". For instance try:

``` LaTeX
\documentclass{article}

\newcounter{theorem}

\begin{document}
\refstepcounter{theorem}
\thetheorem
\end{document}
```

LaTeX has no problem but plasTeX issues a warning `WARNING: unrecognized command/environment: theorem` and fails to output anything in xhtml rendering. The same code with a counter name that doesn't starts with "the" works ok.

kesmit13:
This appears to be fixed now.  Let me know if you see otherwise.

PatrickMassot:
It works for me. Thanks!

Issue #21, opened 2016-01-19
CONVERT ERROR


musashi007:
This is the error I got.

C:\Users\cbt_ones\Desktop>C:\Python27\python.exe C:\Python27\Scripts\plastex sam
ple.tex
plasTeX version 1.0
ERROR: Error while expanding "def" in C:/Users/cbt_ones/Desktop/sample.tex
   on line 2 ('ascii' codec can't encode character u'\ufffd' in position 0:
   ordinal not in range(128))
ERROR: An error occurred while building the document object in
   C:/Users/cbt_ones/Desktop/sample.tex on line 2 ('ascii' codec can't encode
   character u'\ufffd' in position 0: ordinal not in range(128))
Traceback (most recent call last):
  File "C:\Python27\Scripts\plastex", line 113, in <module>
    main(sys.argv)
  File "C:\Python27\Scripts\plastex", line 58, in main
    tex.parse()
  File "C:\Python27\lib\site-packages\plasTeX\TeX.py", line 426, in parse
    for item in tokens:
  File "C:\Python27\lib\site-packages\plasTeX\TeX.py", line 46, in next
    return self._next()
  File "C:\Python27\lib\site-packages\plasTeX\TeX.py", line 335, in __iter__
    tokens = obj.invoke(self)
  File "C:\Python27\lib\site-packages\plasTeX\Base\TeX\Primitives.py", line 159,
 in invoke
    self.ownerDocument.context.newdef(a['name'], a['args'], a['definition'], loc
al=self.local)
  File "C:\Python27\lib\site-packages\plasTeX\Context.py", line 1033, in newdef
    name = str(name)
UnicodeEncodeError: 'ascii' codec can't encode character u'\ufffd' in position 0
: ordinal not in range(128)

In my file, I have this characters.
%\paragraph{配点}
\def\配点pi{5}
\def\配点pii{5}
\def\配点piii{15}
\def\配点piv{0}
\def\配点pv{0}
\calcval[d]{\配点pi+\配点pii+\配点piii+\配点piv+\配点pv}\配点p

tiarno:
does your latex file compile with pdflatex?

musashi007:
I tried doing platex sample2e.tex to check the file and this is the output. 

C:\Users\cbt_ones\Desktop>platex sample2e.tex
This is pdfTeX, Version 3.1415926-1.40.11 (MiKTeX 2.9)
platex: The memory dump file could not be found.
platex: Data: platex.fmt

[...]

Issue #39, opened 2016-08-04
Black square before sign "registered trademark"  (\textregistered)


mdk73:
plasTeX renders a black square before the sign "registered trademark"  (\textregistered).

My test code:

``` LaTeX
\documentclass[a4paper,10pt,fleqn, twoside]{book}
\setlength{\textwidth}{15cm}
\setlength{\textheight}{23.5cm}
\setlength{\evensidemargin}{0cm}

%according Plastex reference section 4.1
\newif\ifplastex
\plastexfalse

\usepackage[bookmarks=true]{hyperref}

\ifplastex
%do nothing
\else
\hypersetup{
    unicode=true,          % non-Latin characters in Acrobat’s bookmarks
    pdffitwindow=false,     % window fit to page when opened
    pdfstartview={FitV},    % fits the height of the page to the window
    colorlinks=true,       % false: boxed links; true: colored links
    linkcolor=blue,          % color of internal links
    citecolor=green,        % color of links to bibliography
    filecolor=blue,      % color of file links
    urlcolor=blue           % color of external links
}
\fi

\begin{document}

%\chapter{Get the current Year}
Trademark\raisebox{1ex}{\tiny{\textregistered}}.

\end{document}
```

The shell output:

```
C:\Users\krause\workspace\manualUSB_en\plastex_test\trademark>c:\python27\python.exe c:\python27\Scripts\plastex --index
-columns=1 "--title=test case" .\plastex_trademarks.tex
plasTeX version 1.0
 ( c:\python27\lib\site-packages\plasTeX\Packages\book.pyc )....
 ( c:\python27\lib\site-packages\plasTeX\Packages\hyperref.pyc )......
INFO: Directing output files to directory: plastex_trademarks.
INFO: Importing templates from c:\python27\lib\site-
   packages\plasTeX\Renderers\PageTemplate
INFO: Importing templates from c:\python27\lib\site-
   packages\plasTeX\Renderers\XHTML
INFO: Importing templates from c:\python27\lib\site-
   packages\plasTeX\Renderers\XHTML\Themes\default
INFO: Using the imager "dvipng".
 [ index.html ]
WARNING: PIL (Python Imaging Library) is not installed.  Images will not be
   cropped.
Traceback (most recent call last):
  File "c:\python27\lib\site-packages\plasTeX\Imagers\__init__.py", line 737, in convert
    dest.crop()
  File "c:\python27\lib\site-packages\plasTeX\Imagers\__init__.py", line 265, in crop
[...]

Issue #29, opened 2016-06-27
Naive newtheorem implementation


PatrickMassot:
Although I don't understand much of plasTeX's internals, I tried my luck at implementing a useful `\newtheorem` command (as far as I understand, the current version of plasTeX is instructed not to complain about it, but not do anything either). 

In order to be able to use (subclasses of) the PageTemplate renderer, I'd like all environment to appear with the same `nodeName` in DOM but with some class attribute recording the type of theorem, so that the template can set some CSS class using it. For instance my current template collection contains only one template for everything created using newtheorem (using jinja2 templating but the meaning should be clear for anyone):

``` html
name: thmenv
<div class="{{ obj.thmName }}_wrapper">
  <span class="{{ obj.thmName }}_heading">{{ obj.caption }}</span> 
  <span class="{{obj.thmName}}_label">{{obj.ref}}</span>
  <div class="{{ obj.thmName }}_content">
  {{ obj }}
  </div>
</div>
```

and it's then up to CSS to distinguish between theorems, lemmas, corollaries, etc.

Here my naive attempt (somewhat cluttered by Issue #28, sorry):

``` python
class newtheorem(Command):
    args = 'name:str [ counter:str ] caption [ within:str ]'
    def invoke(self, tex):
        self.parse(tex)
        attrs = self.attributes
        name = attrs['name']
        counter = attrs['counter']
        caption = attrs['caption']
        within = attrs['within']
        if not counter:
            # Try to avoid newcounter bug when counter name starts with "the"
            if name.startswith('the'):
                counter = 'cnt' + name
            else:
                counter = name
            if within:
                self.ownerDocument.context.newcounter(counter,initial=1,resetby=within)
            else:
                self.ownerDocument.context.newcounter(counter,initial=1)
        # Try to avoid newcounter bug when counter name starts with "the"
        elif counter.startswith('the'):
            counter = 'cnt' + counter
        deflog.debug('newtheorem %s', name)

        # The nodeName key below ensure all theorem type will call the same
        # rendering method, the type of theorem being retained in the thmName
        # attribute
        newclass = new.classobj(str(name), (Environment,), 
                {'caption': caption, 'nodeName': 'thmenv', 'thmName': name,
                    'counter': counter})
        self.ownerDocument.context.addGlobal(name, newclass)
```

It does creates nodes that are intercepted by my template and correctly styled by CSS. But the counter thing does not work at all. If my preamble contains

``` latex
\newtheorem{thm}{Theorem}
[...]

PR #43, opened 2016-08-19
Fix pdf documentation generation

PatrickMassot:
As I mentioned in #27, I wasn't able to compile the documentation. Today I decided to tackle this problem. I identified a couple of problems that prevented my TeXLive 2015 to compile it.

First the \code command is commented out in [python.sty](https://github.com/tiarno/plastex/blob/116356ca6253aec3d2afbdf3fd1e095618cb521e/Doc/texinputs/python.sty#L847), and actually the next line doesn't make any sense without that one.

Instead of using the makeidx package (or one of its many replacements), [plastex.tex](https://github.com/tiarno/plastex/blob/116356ca6253aec3d2afbdf3fd1e095618cb521e/Doc/plastex.tex#L72) tries to directly input plastex.ind. This file doesn't exist before running LaTeX at least once.

There is some strange conflict with ipdf. pdfLaTeX complains this is defined twice.

And I corrected two mistakes in renderers.tex. In #27, I forgot  to add a column and I inserted a line at the wrong place. I didn't see it because I wasn't able to compile.

With this PR I can now compile using both pdfLaTeX and plasTeX.

### DIFF
--- a/Doc/plastex.tex
+++ b/Doc/plastex.tex
@@ -2,6 +2,7 @@
 
 \usepackage[T1]{fontenc}
 \usepackage{graphicx}
+\usepackage{makeidx}
 \usepackage{hyperref}
 
 \newcommand{\titleref}{\ref}
@@ -69,6 +70,6 @@
 
 \input{parsing-faq}
 
-\input{plastex.ind}
+\printindex
 
 \end{document}
--- a/Doc/renderers.tex
+++ b/Doc/renderers.tex
@@ -600,8 +600,8 @@ names in the document object.
 The extensions used for all templating engines are shown in the table below.
 \begin{tableiii}{l|l|l}{}{Engine}{Extension}{Output Type}
 \lineiii{ZPT}{.html, .htm, .zpt}{HTML}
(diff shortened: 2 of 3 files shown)


kesmit13:
Sorry Tim and I haven't been much help lately.  We're both buried at work right now.

PatrickMassot:
There is no problem. I was rather active these past two weeks because I'm on vacations. There are no really blocking issues for me at the moment, although it would be nice to fix #28 and #29. I think it would also be a good idea to close a couple of issues where there is no real problem or no developer can reproduce the problem, and the discussion stopped (I would close #21, #23,  #24, #39 and probably #22). It would clarify things and help gathering motivation to tackle some real issues.

Most of my efforts went to my html5 branch (as discussed in #26). It's almost done. Next week I need to type a lot of lecture notes targeting both pdf and html. If I'm happy with the result I will probably submit a PR around September 1st.

Then I'll try to figure out whether I can help with the python3 branch that Tim started. I'm not sure I will have much time tough, it will depend on external constraints.

Issue #2, opened 2014-05-21
problem with \hyperref [label] {text}


Echeban:
See example below:
\documentclass{article}
\usepackage{hyperref}
\begin{document}
\section{Ordinary Text}
\label{label1} See also \hyperref[label2]{Displayed Text}. 
This works fine in the PDF version but it does not in the HTML version. I was expecting the words `Displayed Text' to provide a link to the section 2. But it does not happen. 
\section{Displayed Text}
\label{label2}
Lorem ipsum. Lorem ipsum. Lorem ipsum. 
\end{document}

Issue #1, opened 2014-04-10
Python 3 Support


Luthaf:
Is there any support planned for Python 3 ? 

PlasTeX is awesome, but it is the only lib preventing our project from using Python 3. And Python 3 is great, even considering only UTF8 as default encoding !

It looks like [this fork](https://github.com/NextThought/nti.plasTeX) is trying to add support. 

What is your position about this ?

PR #46, opened 2016-12-06
Partial support for `amsthm.sty` LaTeX Package

anteprandium:
From the docstring:

plasTeX.Package.amsthm.py: Partial support for `amsthm.sty` LaTeX Package.

This python package provides support for

* Command: `\newtheorem`
* Command: `\newtheoremstyle`
* Command: `\theoremstyle`
* Environment: `proof`

Keep in mind that when you define and use a new theorem-like environment:

    \newtheorem{theorem}{Theorem}
    \begin{theorem}[Euclid's] This is a Theorem. \end{theorem}

this package translates the above construction to simplified LaTeX code (as opposed to a `theorem` environment that can be styled throught the renderer.)

I chose this instead of relying on the renderer because the user can specify everything, from the names to the header fonts, to the colors, to the body font, etc. I don't think the rendering mechanism can cope with environments of user defined names and user defined styles.

It supports the starred version (without numbers) of `\newtheorem`

This package keeps a list of user defined styles in the `userdata` space under `packages/amsthm/styles`. This userdata is prepopulated with (a simplified version of) the  default styles in amsthm.sty: plain, remark, and definition.

It also keeps a list of theorem-like environments under `packages/amsthm/theorems`.

Detailed support:

1. `\newtheorem * {name}[shared counter]{Header}[reset by]` is supported. It creates an environment `\begin{name}[optional note] .. \end{name}`, which translates on invocation into simpler LaTeX.

2. `\newtheoremstyle {name} {above space} {below space} {main body font} {indent amount} {header font} {punctuation hafter head} {space between head and body} {header spec}` is partially supported---it currently ignores everything but main body font and header font. The header spec is hardcoded and equivalent to `\thmname{#1}\thmnumber{ #2}\thmnote{ (3)}`. Read the `amsthm` manual for more information on defining theorem styles. There's a list of defined styles in `userdata` space, in `packages/amsthm/styles`.

3. `\theoremstyle{name}` Set the current theorem style to `name`. Not checked for errors.

4 `\begin{proof} .. \end{proof}` Translated into simpler LaTeX.

5. `\swapnumbers`  Ignored & Unsupported.

TODO & NOT FULLY WORKING:

1. I can't get `\label` to  reference properly the correct statement.

2. For some reason, I can't get

        \newtheorem{theorem}{Theorem}
        \begin{theorem} Statement \end{theorem}

    to wrap the Statement in a span. Currently it gives you something like:

        <p><b>Theorem 1</b> <i>Statement</i><p>

    and I would like something like:

        <span class="statement"><p><b>Theorem 1</b> <i>Statement</i><p></span>

3. Support of `header spec`.

### DIFF
--- a/plasTeX/Packages/amsthm.py
+++ b/plasTeX/Packages/amsthm.py
@@ -1,0 +1,285 @@
+#!/usr/bin/env python
+
+"""
+
+plasTeX.Package.amsthm.py: Partial support for `amsthm.sty` LaTeX Package.
+
+This python package provides support for
+
+
+* Command: `\newtheorem`
+* Command: `\newtheoremstyle`
+* Command: `\theoremstyle`
+* Environment: `proof`
+
+Keep in mind that when you define and use a new theorem-like environment:
+
+    \newtheorem{theorem}{Theorem}
+    \begin{theorem}[Euclid's] This is a Theorem. \end{theorem}
+
+this package translates the above construction to simplified LaTeX code (as opposed to a `theorem` environment that can be styled throught the renderer.)
+
+I chose this instead of relying on the renderer because the user can specify everything, from the names to the header fonts, to the colors, to the body font, etc. I don't think the rendering mechanism can cope with environments of user defined names and user defined styles.
+
+It supports the starred version (without numbers) of `\newtheorem`
+
+This package keeps a list of user defined styles in the `userdata` space under `packages/amsthm/styles`. This userdata is prepopulated with (a simplified version of) the  default styles in amsthm.sty: plain, remark, and definition.
+
+It also keeps a list of theorem-like environments under `packages/amsthm/theorems`.
+
+Detailed support:
+
+1. `\newtheorem * {name}[shared counter]{Header}[reset by]` is supported. It creates an environment `\begin{name}[optional note] .. \end{name}`, which translates on invocation into simpler LaTeX.
+
+2. `\newtheoremstyle {name} {above space} {below space} {main body font} {indent amount} {header font} {punctuation hafter head} {space between head and body} {header spec}` is partially supported---it currently ignores everything but main body font and header font. The header spec is hardcoded and equivalent to `\thmname{#1}\thmnumber{ #2}\thmnote{ (3)}`. Read the `amsthm` manual for more information on defining theorem styles. There's a list of defined styles in `userdata` space, in `packages/amsthm/styles`.
+
+3. `\theoremstyle{name}` Set the current theorem style to `name`. Not checked for errors.
+
+4 `\begin{proof} .. \end{proof}` Translated into simpler LaTeX.
+
+5. `\swapnumbers`  Ignored & Unsupported.
+
+TODO & NOT FULLY WORKING:
+
+1. I can't get `\label` to  reference properly the correct statement.
+
+
+2. For some reason, I can't get
+
+        \newtheorem{theorem}{Theorem}
+        \begin{theorem} Statement \end{theorem}
+
+    to wrap the Statement in a span. Currently it gives you something like:
+
+        <p><b>Theorem 1</b> <i>Statement</i><p>
+
+    and I would like something like:
+
+        <span class="statement"><p><b>Theorem 1</b> <i>Statement</i><p></span>
+
+3. Support of `header spec`.
+
+
+Example document:
+
+    \documentclass{article}
+    \usepackage{amsthm}
+    \usepackage{color}
+
+    \newtheoremstyle{redPlain}  {\bigskipamount}
+                                {\bigskipamount}
+                                {\itshape}
+                                {}
+                                {\color[cmyk]{0,1.00,0.65,0.34}\bfseries}
+                                {:}
+                                {1em}
+                                {\thmname{#1}\thmnumber{ #2}\thmnote{ (#3)}}
+    \theoremstyle{redPlain}
+    \newtheorem*{starred}{Starred}
+    \newtheorem{normal}{Normal}
+    \newtheorem{normalshared}[normal]{Normal And Shared}
+    \theoremstyle{plain}
+    \newtheorem{within}{Within}[section]
+    \newtheorem{withinshared}[within]{Within And Shared}
+    \begin{document}
+
+    \section{A section}
+    \begin{starred}[Note]
+        An starred theorem
+    \end{starred}
+    \begin{normal}
+        A normal theorem
+    \end{normal}
+    \begin{normalshared}[Note]
+        A normal theorem, with shared counter.
+    \end{normalshared}
+    \begin{within}
+        A theorem numbered within sections
+    \end{within}
+    \begin{withinshared}[Note]
+        A theorem with counter shared with a theorem numbered within sections
+    \end{withinshared}
+    \section{Another section}
+    \begin{starred}
+        An starred theorem
+    \end{starred}
+    \begin{normal}
+        A normal theorem
+    \end{normal}
+    \begin{normalshared}
+        A normal theorem, with shared counter.
+    \end{normalshared}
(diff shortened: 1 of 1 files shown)


anteprandium:
Forgot to mention in the commit comment: the docstring includes a test document, where you can see the styles in action. It also includes colors in the theorem headers!

PatrickMassot:
Thank you very much for this effort. I'm very happy to see I'm not the only one interested in theorems in plasTeX!

For the sake of discussion, I just pushed my html5 branch to github. It's not yet ready for a PR. I need to find a couple of days for polishing stuff and cleaning the history. This clearly won't happen before Christmas break. See however [my newtheorem](https://github.com/PatrickMassot/plastex/blob/ebb61c52ad73b96ed7c22b2b9955f5a68db3f2f5/plasTeX/Base/LaTeX/Definitions.py#L86) and the [corresponding jinja2 template](https://github.com/PatrickMassot/plastex/blob/ebb61c52ad73b96ed7c22b2b9955f5a68db3f2f5/plasTeX/Renderers/HTML5/Thms.jinja2s). 

I have no time to read your code today but it sounds like we could get something really nice by merging somehow our efforts. In particular my theorem environments do generate an environment which is styled by the renderer only. See my [lectures notes](http://www.math.polytechnique.fr/perso/massot.patrick/enseignement/mat553/sect0012.html) generated by plasTeX, paying attention both to the html markup and the css files [coming from theme](http://www.math.polytechnique.fr/perso/massot.patrick/enseignement/mat553/styles/theme-style.css) and [custom to this project](http://www.math.polytechnique.fr/perso/massot.patrick/enseignement/mat553/styles/stylecours.css).

anteprandium:
@PatrickMassot  I took a quick look at your code, it's way cleaner than mine, and I've seen several tricks worth knowing, so thanks a lot for sharing your code!

There's a difficult-to-balance situation here: the `amsthm`package lets you choose, say, green theorems with bold header and italic body, and blue remarks, with italic header and regular body font. The renderer would have to be really clever to handle both! This is an extreme example, and a sure ugly setup, but you get the idea.

On the other hand, one could consider the theorems environments as “logical markup” and output a more or less fixed rendering, without acknowledging user-defined styling.

It seems we took different sides on which to lean on :) 

Maybe we can work out a solution over Christmas break, on n'est pas tellement pressé!!!

Thanks again, in any case.

anteprandium:
Forgot to mention: your notes are _absolutely_ beautiful!

anteprandium:
@PatrickMassot I'll address here the documentation part, and in a different comment the renderer styling of theorems.

What I've found most difficult to understand (I am not sure I have it completely figured out) is the read-expand-evaluation loop. For one thing, it took me ages to find out about `tex.input()`, which is fantastic for doing stuff in python, and injecting back code into the stream.

Another big bump was to realise that the evaluation of `Environment`s is different from LaTeX: in LaTeX, any environment is actually composed of two commands: `\begin{env} .. \end{env}` is `\env ... \endenv` plus some grouping and book-keeping. Here the same command is called twice, with different modes. 

As for the tricks in your code:

1.  `nodeName`! Once you see it, it is obvious, but until then....

2. `obj.ref`, `obj.id`. This is (I think) how labelling and referencing work, but I couldn't for the life of me figure this one out.

3. I assume the `proof.digest` environment is worth investigating. 

So, perhaps the documentation could offer an nontrivial addition to `Command`s and `Environment`s, which uses `invoke`, `digest` _et al_, as an example for developers? Maybe the very same `\newtheorem`, as a *Developer example* section? 

Just a thought.

anteprandium:
@PatrickMassot The example I have in mind is the use of `\newtheoremstyle`, as per `amsthm` package. Consider this valid document:

    \documentclass{article}
    \usepackage{amsthm}
    \newtheoremstyle{awfulBlue}
                                  {\bigskipamount} % Space above
                                  {\bigskipamount} % Space below
                                  {\normalfont\color[named]{Blue}\itshape} % Main body font.
                                  {0pt} % Indentation
                                  {\normafont\bfseries\color[named]{yellow}} % Header font
                                  {:} % Punctuation after header
                                  {1.5em} % Space between head and body
                                  {\thmname{#1}\thmnumber{ #2}\thmnote{#3}} % Space above
    \newtheoremstyle{perkyRed}
                                  {\bigskipamount} % Space above
                                  {\bigskipamount} % Space below
                                  {\normalfont\color[named]{red}\upshape} % Main body font.
                                  {0pt} % Indentation
                                  {\normafont\itshape\color[named]{red}} % Header font
                                  {.} % Punctuation after header
                                  {1.5em} % Space between head and body
                                  {\thmname{#1}\thmnumber{ #2}\thmnote{#3}} % Space above

    \theoremstyle{awfulBlue}
    \newtheorem{theorem}{Theorem}
    \theoremstyle{perkyRed}
    \newtheorem{remark}{Remark}

    \begin{document}
    \begin{theorem}
        This is an awful blue Theorem; blue, italic body and yellow bold header. Header and body are separated by a ":".
    \end{theorem}
    
    \begin{remark}
	 This is a perky red Remark; red upshape body and red italic header. Header and body are separated by a ".".
    \end{remark}
    \end{document}

So you see, style and content **are** separated, but the document can well choose the styling with `\newtheoremstyle` and `\theoremstyle`. If these two are to be supported, the style has to go from the document to the renderer. Now, the point is: any (valid) LaTeX construct can go into the header and body part. This is way too general for the renderer to handle, I think, and way too late in plasTeX processing to allow major changes.

That's why I commited to `amsthm.py` and not to `Definitions.py`. The `amsthm.sty` package is much more ambitious than LaTeX's default theorem handling. Maybe I am too ambitious myself?

As a final note: I am conscious that my code is not up to par. I just hoped to spark the thought process of how to support `amsthm.sty`, by offering _something_, that could (surely!) be improved.

Thanks again, and don't let the exam leak online ;) (Though I'll freely admit it'd make no difference to me: I am differential-topology-challenged.)

PatrickMassot:
Thanks for your explanations. I think trying to get plasTeX to implement ``\newtheoremstyle`` may be a bad idea. This command is too much tied to pdf output. I don't think there is any hope to get a nice web page produced from exactly the same tex as a nice pdf. For the sake of discussion, I just pushed a slightly simplified version of [my setup](https://github.com/PatrickMassot/mat553) (without the exam). 

Please have a look at [poly.tex](https://github.com/PatrickMassot/mat553/blob/master/src/poly.tex) which is compiled by LuaLaTeX or XeLaTeX and produces [the pdf](http://www.math.polytechnique.fr/perso/massot.patrick/enseignement/mat553/poly.pdf) vs [web.tex](https://github.com/PatrickMassot/mat553/blob/master/src/web.tex) which produces the web pages you've already seen. They both input ``cours.tex`` (which in turn inputs the individual chapters and exercises sets) and  ``macros_common.tex``. But they don't have the same document class and header and they each add one specific macros file. In particular both [macros_poly.tex](https://github.com/PatrickMassot/mat553/blob/master/src/macros_poly.tex) and [macros_web.tex](https://github.com/PatrickMassot/mat553/blob/master/src/macros_web.tex) start with theorems definition using either thmtools (much worse than amsthm from a plasTeX perspective) or my ``\newtheorem`` semi-implementation. Also some LaTeX commands do nothing in one version, especially the ones that generate links between theorems or theorems and exercises (``\uses`` and ``\covers``) do nothing on LaTeX side.

anteprandium:
> This command is too much tied to pdf output. I don't think there is any hope to get a nice web page produced from exactly the same tex as a nice pdf

One can dream :)  But the closer we get there, the better. Of course, some TeX parts, like “Space above“ have no real counterpart in HTML. But others, like colors and fonts, do. This PR was intended as a proof-of-concept.

Still, it works for my use-case, so I might as well keep it. 

Thanks for your time.

PR #57, opened 2016-12-23
Allow macros to define post-parse callbacks

PatrickMassot:
The goal is to allow python packages to interact with plasTeX after parsing (in particular all references are then resolved) but before rendering. This is done by putting call back functions in the document ``post_parse_cb`` list.

I can provide a detailed use case if needed.


PatrickMassot:
Python style guide recommends consistency first so I should have used camelCase. This is modified now.

Issue #61, opened 2016-12-28
Status of DocBook Renderer


PatrickMassot:
In the documentation, there is a section on DocBook Renderer containing only the sentence "Not yet implemented'. But there is a DocBook Renderer in the codebase. Is it a stub or is it usable? I don't know anything about DocBook so I cannot test this renderer but it would be nice to document it, if it is usable.

PatrickMassot:
@tiarno would you have time to write a couple of sentences about this? I've tried again to use the DocBook renderer, but cannot render any TeX file. Do you have any example that compiles using the plastex command? Or is it usable only with plasTeX used as a library?

tiarno:
Hi Patrick, I have never used the plastex command to create DocBook documents, but always used plastex as a library. This is the pattern I use at work. Assumes `fname` is the name of the LaTeX source document and that TEXINPUTS and DocBookRENDERER env vars are set. This will render the xml file in the current directory. 

    from plasTeX.ConfigManager import ConfigManager
    from plasTeX.TeX import TeX
    from plasTeX.Renderers.DocBook import Renderer
    from plasTeX import TeXDocument

    texdocument = TeXDocument(config= ConfigManager())
    tex = TeX(texdocument, myfile=fname)
    tex.parse()
    r = Renderer()
    r.render(texdocument)

PatrickMassot:
Fixed in 4d4dd32.

PR #62, opened 2016-12-29
Html5

### LINKED ISSUE #26: Html5
Are there any short term plans to add an Html5 renderer? I mean a renderer that would output html5+css3 of course, but also use [MathJax](https://github.com/mathjax/MathJax) (or [KaTeX](https://github.com/Khan/KaTeX) maybe).

I'm about to start writing some math lecture notes using PlasTeX and Html5 is an absolute requirement. I'm willing to try writing the renderer (mostly templates actually) but I don't want to duplicate efforts if someone is already working on it (or has already done it somewhere).

PatrickMassot:
Fixes #26.

This is my long overdue HTML5 pull request. The main goal is to have a HTML5+CSS3+JS renderer with a responsive default layout and MathJax handling of mathematics.

A secondary goal was to allow output customization through CSS. Hence templates contain almost no inline styles. Together with the ``--extra-css`` command line option, this allows users to tremendously modify the output design without changing any template. A side effect of this is the creation of an extra layer of customization inside each theme: the theme css layer, controlled by option ``--theme-css`` which only selects a different css file. I could replace this extra layer by a set of themes but it would imply a lot of file duplication.
Currently there are two such theme css: the default one named ``green`` tries to use plasTeX color scheme while ``blue`` mostly uses blue, white and gray. They are both generated using the SASS preprocessor but the built css are included so SASS is not required to use this renderer.

Another secondary goal came from my attempts to develop plasTeX packages for my lecture notes. I wrote packages to handle TikZ pictures and packages to add extra meta-informations like covering reports (analogous to unitests except that here exercices are covering definitions and theorems) and informations about which theorems are used to prove what. Those packages are mostly not included in this PR, except probably some leaks in CSS and JS. I will make PR for them in the near future, either because they are widely useful (TikZ case) or at least because they provide nice examples for package developers. However all these packages needed more hooks from the Renderer (on top of the parser hook implemented in PR #57 and various other recent PR ). Hence this renderer allow packages to override templates, add css and javascript files and register filters.

All this is documented at the end of Chapters 2 and 5.

Note that merging this PR (probably after a number of adjustments) would actually be the minimal route. I tried to minimize the impact of this PR outside of the HTML5 renderer. But one could instead move part of the code to the ``PageTemplate`` renderer (everything which makes no reference to HTML or CSS, like what I added to the ``processFileContent`` and ``cleanup`` methods) and also, probably less useful, backport some code to the ``XHTML`` renderer. I think this would be backward compatible but would probably slow down those renderers a bit. Next one could make HTML5 plasTeX's default renderer. Obviously this would not be backward compatible (and would require documentation change). Also it would make plasTeX depend on Jinja2 since all HTML5 templates are written using this template engine.


PatrickMassot:
For convenience, I compiled the documentation using this branch and published it [there](http://patrickmassot.github.io/plastex/plastex/index.html). In particular the documentation of the HTML5 renderer is [here](http://patrickmassot.github.io/plastex/plastex/sec-config-html5.html) and [there.](http://patrickmassot.github.io/plastex/plastex/sec-html5.html)

PatrickMassot:
I agree it may require more thinking. One easy modification would be to put all my userdata keys in one  dictionary ``userdata['html5']`` (so ``userdata[’processFileContents’]`` would become ``userdata['html5'][’processFileContents’]`` for instance).

About your second paragraph, would you like me to try an implementation of what I understand or to you want to think more?

kesmit13:
I almost think another field should be added to the document so that we aren’t using ‘userdata’.  That really is intended for users.  This feature is for package writers and works directly in conjunction with the renderers, so it's more core functionality.  Adding a layer in the keys that contains the renderer name sounds like a good idea as well.

As far as the attachment types, I was thinking of something like the nav structure in Base/LaTeX/Sectioning.py that uses the HTML link types to point to various sections of the document and other resources.   The first keys we would use would be 'stylesheet' and 'script'.  Inside that, we would need a path to the resource and a mime-type.  The mime type could be used to specify different variants of CSS like SASS and Less so that the renderer would know how to handle it.

I'm mostly just thinking out loud here, but if you want to take a crack at it, I wouldn't mind.

PatrickMassot:
I just tried something new here. I attempted to better separate renderer data from user data and separating renderers by adding, next to ``document.userdata``, a ``document.rendererdata`` which is a dictionary having renderer names as keys. I also tried to use a more generic approach towards adding keys to this dictionary and copying extra files or templates, see the new ``PackageResource`` module. 

For instance, the new way a package can add a css file to be used by the HTML5 renderer is by importing the ``PackageCss`` class from the new ``PackageResource`` module and using in ProcessOptions:
```python
css = PackageCss(
            renderers=['html5'],
            package='my_package',
            data='my_shiny_package.css')
document.packageResources.append(css)
```

While the new way to add a ``preCleanUp`` callback is to use:
```python
cb = PackageResource(
                renderers=['html5'],
                key='preCleanupCallbacks',
                data=myCallBack)
document.packageResources.append(cb)
```

Is it better?

kesmit13:
That's a lot better.  Is the reason you want to use an object so that it can do the actual work of copying files, running callbacks, etc.?   One thing I would probably change would be to add a document.addPackageResource(...) method rather than modifying the instance variable directly.  Also, having a glob or regex pattern for the renderers would be nice so that you could say renderers='*html*'.  Other than that, I think it will work fine.

PatrickMassot:
Initially I wanted the object to do most of the work but I kind of stopped in the middle, as you can see in the code. But the main goal was to avoid changing the renderer code for each new type of resource.
I wanted to avoid things like:
```python
for resource in resources:
  if resource.type == 'css':
    do CSS stuff
  elif resource.type == 'js':
    do Javascript stuf
  elif ...
```
(or having a separate loop for each type of resource as in my previous code) but rather have
```python
for resource in resources:
  resource.doStuff()
```
and each new type of resource becomes a subclass of ``PackageResource``.

I also thought about using a ``document.addPackageResource`` but was unsure whether it was a good idea. I'll do it since you like it.

About the regex pattern for renderers, how would you indicate that a resource is useful for, say 'xhtml', 'html5' and 'epub'? 

Globally I wanted to have your opinion before pushing this idea further but I still need to work a bit on all this.

kesmit13:
As far as using a regex for matching renderers, I would suggest adding a method to the PackageResource class that would walk through the renderers list and do a regex match on each one to see if it matches a given input string (which contains the name of the renderer).

Once you add that (and any other things you may have thought of), I'm fine with this pull request.

PatrickMassot:
I implemented this regex change and also completed moving all the logic into the PackageResource class. Tell me if you like it because I would need to update the documentation before you could merge.

kesmit13:
This looks fine to me.

PatrickMassot:
Ok I'll probably write documentation tonight. 

Do you think HTML5 should become the default renderer?

kesmit13:
I'll have to let @tiarno decide that since he is the primary user in our company.  I think he may use a DocBook renderer though.  My gut feeling at the moment is that maybe we make it the default in the Python 3 branch, or maybe we do a plasTeX 2.0 release and do it there.

PatrickMassot:
Changing the default renderer would of course need a new release. It probably makes sense to have a python2 release with the new renderer but keeping XHTML default and then make a python3 release with HTML5 default. 

Starting a 2.x numbering with the python 3 release is also rather natural. But we're not there yet. I still can't compile plasTeX documentation with the python3 branch and then the next goal would be to incorporate all 58 commits + HTML5 from master into python3...

kesmit13:
I'm still hopeful that once we get Python 3 working we might be able to make that branch work with Python 2 as well.  If so, we'll just release it as v2 and make HTML5 the default for both.

PatrickMassot:
That would be really nice but I don't think it will be easy. Tim had to work quite a lot to get to the python3 branch.

tiarno:
Hi, I don't have any emotional connection to the python3 branch I started. When I started it, it was just to see if I could get it to work for myself since no one else was interested in it. There was some work involved, but I left a lot of stuff 2to3 put it that probably isn't necessary (not every dictionary view needs to be coerced to a list). 
As for the renderer, I do use the DocBook renderer at work; I never use the XHTML renderer any more. So HTML5 would be fine with me since it won't affect me--I subclass from the PageTemplate Renderer.

Work has been blowing up for months now, it seems like the pace has just increased and will never slow down again. I appreciate the work you guys are doing. I would love to have a stable python3 branch.

kesmit13:
I'm the same way at work.  I try to get in and review code whenever I get a change, but getting time to actually write code is tough.

PatrickMassot:
I'll probably manage to find time to bring the python3 branch up to date. My trick is that writing software is not my job so it qualifies as leisure, although my mind is not always fit for software writing after a day at work.

Tim, if you can find some time to work on plastex, it would be nice to tackle issue #61 (document the DocBook renderer) since you say you use it. The current documentation about this renderer is: "Not implemented".

PatrickMassot:
This is hopefully the last bunch of commits on this PR. I added the possibility to put extra files in user directory instead of plasTeX main library. Indeed, one can add python packages in the directory where plasTeX is invoked (something I hopefully clarified in 51782ec) so there is no reason such packages could not use the PackageResource stuff. Then I rewrote the documentation to explain this new way of doing things.

Since we are also working on tests to help with the python3 transition, I wrote unit tests covering the new PackageResource module as well as the corresponding functional tests. They both need pytest though.

Doc/renderers.tex
```diff
@@ -1308,43 +1308,57 @@ \section{HTML5 Renderer}\label{sec:html5}
 
 \subsection{Interactions with packages}\label{sec:html5-pkg}
 
-This section is useful for packages writers.  The HTML5 Renderer allows
-packages to interact with the rendering process in several ways. 
+This section is useful for packages authors. The HTML5 Renderer allows
+packages to interact with the rendering process in several ways through
+packages resources. Each such resources is wrapped in an object whose
```
kesmit13: "resource" should be singular.

Doc/renderers.tex
```diff
@@ -1308,43 +1308,57 @@ \section{HTML5 Renderer}\label{sec:html5}
...
-be loaded (at least when using the default layout), in alphabetical
-order, after the theme Javascript files, but before user defined extra
-Javascript files.
-For instance, if a package \file{mypackage.py} wants to use all three
-possibilities, it must create, inside \file{plasTeX/Renderers/HTML5/}
-subdirectories \file{pkgTemplates/mypackage},
-\file{pkgCss/mypackage}, \file{pkgJs/mypackage} and contain a
-\function{ProcessOptions} function including the following:
+Inside the main renderer directory \file{plasTeX/Renderers/HTML5/}
+packages authors can add directories containing extra input files.
```
kesmit13: "packages" should be singular.

kesmit13:
Thanks for all of the help on the new renderer.  HTML5 is a nice addition.

PR #75, opened 2017-02-06
Python3 level up

PatrickMassot:
This PR hopefully brings the python3 branch up to date with the master branch.

I'm not an expert at git cherry-pick and git rebase so the process has been a bit chaotic. So the history is not quite what I hoped for. I think I've stolen ownership of a couple of commits, I'm sorry. But I hope the result is mostly ok.

Modifications of TeX.py in 7296136d69a918f3 and cc57edef06a0b are conflicting but seem to do roughly the same so I kept the version in python3 branch, but incorporating the idea of 7dfe9abc40a

Following issue #21 (that could be renamed something like "unicode in macro names"), commits 8bd6fab20ce, 53869cb,  cc57edef06a0b work toward allowing unicode in macro names. But, as far as I can see, it still doesn't work in master, so I didn't try to port this code to python3 branch.


tiarno:
Looks good Patrick, thanks for the changes.

PR #78, opened 2017-02-21
No mutable default values

### LINKED ISSUE #77: Filename generation reset
Output file names generation is driving me crazy. When I try to render two documents in the same python script then it seems available filenames are not reset, although I recreate *all* objects. Say I have inside ``/tmp/test.py``:
```python
import os
from plasTeX.TeX import TeX, TeXDocument
from plasTeX.Config import config
from plasTeX.Renderers.XHTML import Renderer

cur_dir = os.getcwd()
try:
    os.makedirs('build')
except:
    pass

doc = TeXDocument(config=config)
tex = TeX(doc)
tex.input("\\documentclass{article}\\begin{document}Hello.\\end{document}")
doc = tex.parse()
doc.userdata['working-dir'] = '.'

renderer = Renderer()
os.chdir('build')
renderer.render(doc)

os.chdir(cur_dir)
try:
    os.makedirs('build2')
except:
    pass
doc = TeXDocument(config=config)
tex = TeX(doc)
tex.input("\\documentclass{article}\\begin{document}Hello.\\end{document}")
doc = tex.parse()
doc.userdata['working-dir'] = '.'

renderer = Renderer()
os.chdir('build2')
renderer.render(doc)
```
The first rendering renders to ``build/index.html`` but the second one renders to ``build2/sect0001.html``!!!

I don't understand what I'm doing wrong but it is a huge problem when writing unittests. I need my tests to be isolated from each other but, whatever I do, some devil remembers I already rendered some ``index.html``. I tried using ``deepcopy`` everywhere, using ``del`` after using every objects but nothing worked.

PatrickMassot:
Fixes #77 and maybe other unknown bugs.

Remove mutable in function default values except in few cases where the value is copied anyway. This is a systematic change. I'm aware it may introduce other bugs (all tests that used to pass still pass though) so I would understand if you prefer to fix only ``Filenames.py``.

### DIFF
--- a/plasTeX/Renderers/PageTemplate/simpletal/simpleTAL.py
+++ b/plasTeX/Renderers/PageTemplate/simpletal/simpleTAL.py
@@ -830,7 +830,7 @@ class TemplateCompiler:
 		else:
 			self.commandList.append (command)
 		
-	def addTag (self, tag, tagProperties={}):
+	def addTag (self, tag, tagProperties=None):
 		""" Used to add a tag to the stack.  Various properties can be passed in the dictionary
 		    as being information required by the tag.
 		    Currently supported properties are:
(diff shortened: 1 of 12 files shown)


plasTeX/Renderers/PageTemplate/simpletal/simpleTAL.py
```diff
@@ -841,10 +841,11 @@ def addTag (self, tag, tagProperties={}):
 					'singletonTag'    - A boolean to indicate that this is a singleton flag
 		"""
 		# Add the tag to the tagStack (list of tuples (tag, properties, useMacroLocation))
+                tagProperties = tagProperties or {}
```
kesmit13: This looks like an indentation bug.

kesmit13: Maybe... the code in this listing looks like it's indented improperly.  But looking at the full file it looks fine.

kesmit13: I think these changes look fine.  I don't see why it would break anything and it's better to just get this out of the way now to prevent any future issues.

Issue #45, opened 2016-11-08
Preventing character substitution


PatrickMassot:
In ``plasTeX/__init__.py`` there is a list ``charsubs`` of "Character sequences that should be replaced by unicode". Under what circumstances is this list used? It sometimes get into my way but it's not easy to predict.

For instance try putting into ``test.tex`` the only line
```tex
\draw (0,0) -- (1,1) decorate { -- (2,1) } -- (3,0);
```
which could be part of a valid tikzpicture. Of course there is no hope for plasTeX to parse the original pgf/tikz package. The only thing I want plasTeX to do is to expand macros before passing the TeX source to pdflatex and pdf2svg and then include the resulting svg file into my html file.
See e.g. http://www.math.polytechnique.fr/perso/massot.patrick/enseignement/mat553/sect0023.html for a page produced by plasTeX using this strategy.

But the second  --  gets mangled to a unicode dash because of line 803 in the file mentioned above. Not the first one and not the third one, only the one inside the ``bgroup``:
```python
In [1]: from plasTeX.TeX import TeX

In [2]: doc=TeX(file='test.tex').parse()
WARNING: unrecognized command/environment: draw

In [3]: doc.childNodes[7:9]
Out[3]: [-, -]

In [4]: doc.getElementsByTagName('bgroup')[0].childNodes
Out[4]: [u' \u2013 (2,1) ']
```

So questions are:
* what is the rule deciding whether these character substitutions occur?
* is there any way I can tell plasTeX that it shouldn't not do that inside a tikzpicture environment say?

Right now I'm using a version of plasTeX where the offending line 803 is commented out but this is not a good workaround.

kesmit13:
The best way to fix this would be to implement the tikzpicture package and put it in the plasTeX/Packages/ directory.  You don't have to actually implement all of the macros in that environment, but you do need to put placeholders in for them so that plasTeX won't complain about them.  

Then you override the normalize method to prevent it from using the character substitutions.  The class would look something like the following.  Let me know if this works out for you.  It is one of those edge cases that isn't that common.

class tikzpicture(Environment):

    class draw(Command):
        pass

    def normalize(self, charsubs=[]):
        """ Normalize, but don't allow character substitutions """
        return Environment.normalize(self)

PatrickMassot:
Implementing a stub tikz package is already what I did to achieve the result, only needing to comment out that line about ``--``. 

[...]

PR #81, opened 2017-02-22
Tikz support

PatrickMassot:
This PR brings support for tikz and tikz-cd packages when using the HTML5 renderer. This was already discussed quite a bit in issue #45 and PR #62. See also the documentation I wrote.

I guess there should be a way to do that using the Imager mechanism but I had no success at all along this path. There were too many things I didn't understand. The current implementation works for me but feel free to redo it the right way.

Last note: this package and its tests are fully compatible with python3. If this is merged then I'll be able to fill a python3 PR instantly, using only cherry-picked commits.


plasTeX/Packages/tikz.py
```diff
@@ -0,0 +1,133 @@
...
+        object_id = env.attrs['id']
+        basepath = os.path.join(tmp_dir, object_id)
+        texpath = basepath + '.tex'
+        pdfpath = basepath + '.pdf'
+        svgpath =  basepath + '.svg'
+
+        context = {'tikzpicture': env.text.strip() }
+        template.stream(**context).dump(texpath, encoding)
+
+        os.system(compiler + ' ' + texpath)
```
kesmit13: Can these os.system calls be replaced with subprocess.call?  If the paths contain spaces, this could fail.

PatrickMassot: Sure. I'll do that in the evening.

PatrickMassot:
I fixed this issue. I also took the opportunity to add a true functional test actually calling the main plastex script, pdflatex and pdf2svg. It turned out to be a good idea. In my real world examples the output directory was always set and I didn't realize that the default value needed template interpolation. And also the images directory may not exist in a super simple example...

PatrickMassot:
cherry-picked

PR #96, opened 2017-08-12
Amsmath numberwithin

PatrickMassot:
Implements command ``\numberwithin`` in package ``amsmath``.


pbelmans:
There is a small issue with the numbering. Consider the following code:
```latex
\documentclass{article}
\usepackage{amsmath}
\numberwithin{equation}{subsection}

\begin{document}
\begin{equation}
  a^2+b^2=c^2
\end{equation}
\end{document}
```

The number should be 0.0.1, but it is 1.  Adding a section (but no subsection) makes it 1.1 (and it should be 1.0.1), adding a subsection makes it the correct value of 1.1.1.

pbelmans:
A similar issue appears with numbering theorems. If you are numbering them within subsections, but there is no surrounding subsection because you are in a new section without having started a subsection, the numbering should be `7.1.0.1`, but plasTeX generates `7.1.1`.

pbelmans:
This seems to be intended behaviour, according to [this line](https://github.com/tiarno/plastex/blob/7803cc55187ed90ec493e1aa906d2f76efaf3218/plasTeX/__init__.py#L1588).

Issue #93, opened 2017-08-08
Document logging levels


PatrickMassot:
Somewhere in the documentation, in some section intended for people who want to contribute to plasTeX or write a new renderer, we should document ``plasTeX/plasTeXrc``. As far as I understand, this is the only place where changing the log levels has any chance to have any impact. Indeed, the main script gets a logger instance before reading any other configuration file and it seems it is then too late to adjust log levels (but I'd be very happy to be proven wrong).

I realized this is not documented when I wanted to point it out to people involved in the effort to use plasTeX for the new [Stacks project](http://stacks.math.columbia.edu/) website, currently going on [here](https://github.com/pbelmans/plastex/tree/gerby)

PR #98, opened 2017-08-12
Python package for debugging purposes

PatrickMassot:
This is a package that I've been using for a while. But discussions with the stacks project people convinced me I should share it. This package introduces a couple of commands for debugging purposes:
* a command to set logging levels like ``\setloglevel{parse.definitions}{DEBUG}`` (related to issue #93)
* a package argument ``\usepackage[post_parse_trace]{debugplastex}`` which drops into a debugger after parsing.
* a command ``\settrace`` to drop into a debugger during parsing.

I wanted to write documentation for this, but I couldn't decide which chapter to put it.

### DIFF
--- a/plasTeX/Packages/debugplastex.py
+++ b/plasTeX/Packages/debugplastex.py
@@ -1,0 +1,35 @@
+import pdb
+
+from plasTeX import Command
+from plasTeX.Logging import getLogger
+
+
+class settrace(Command):
+    def invoke(self, tex):
+        document = self.ownerDocument
+        config = document.config
+        context = document.context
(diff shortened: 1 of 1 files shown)


kesmit13:
This one is pretty interesting.  I'd never thought of doing that.

PatrickMassot:
Do you have any opinion about where to explain this in the documentation?

kesmit13:
I'd say either add it as a FAQ entry (even though it's not asked that often), or add another appendix.

PatrickMassot:
Ok, I'll do that in the near future.

Issue #109, opened 2019-03-12
Make plastex work with Python 3.7


kodyvajjha:
## Issue
PlasTeX is currently incompatible with python 3.7 due to [changes in exception handling inside generators](https://www.python.org/dev/peps/pep-0479/). 
Here is the traceback:

```
plasTeX version 1.0
ERROR: An error occurred while building the document object in <tokens> on
   line 1 (generator raised StopIteration)
ERROR: Error while reading argument "options" of documentclass in <string>
   on line 1 (generator raised StopIteration)
ERROR: Error while expanding "documentclass" in <string> on line 1
   (generator raised StopIteration)
ERROR: An error occurred while building the document object in <string> on
   line 1 (generator raised StopIteration)
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/plasTeX/Tokenizer.py", line 386, in __iter__
    token = next(charIter)
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/bin/plastex", line 142, in <module>
    main(sys.argv)
  File "/usr/bin/plastex", line 90, in main
    tex.parse()
  File "/usr/lib/python3.7/site-packages/plasTeX/TeX.py", line 435, in parse
    for item in tokens:
  File "/usr/lib/python3.7/site-packages/plasTeX/TeX.py", line 46, in __next__
    return self._next()
  File "/usr/lib/python3.7/site-packages/plasTeX/TeX.py", line 343, in __iter__
    tokens = obj.invoke(self)
  File "/usr/lib/python3.7/site-packages/plasTeX/Base/LaTeX/Packages.py", line 36, in invoke
    a = self.parse(tex)
  File "/usr/lib/python3.7/site-packages/plasTeX/__init__.py", line 472, in parse
    **arg.options)
  File "/usr/lib/python3.7/site-packages/plasTeX/TeX.py", line 757, in readArgumentAndSource
    toks, source = self.readGrouping(spec, expanded, parentNode=parentNode)
  File "/usr/lib/python3.7/site-packages/plasTeX/TeX.py", line 909, in readGrouping
    toks = self.expandTokens(toks, parentNode=parentNode)
  File "/usr/lib/python3.7/site-packages/plasTeX/TeX.py", line 406, in expandTokens
    out = tex.parse(frag)
  File "/usr/lib/python3.7/site-packages/plasTeX/TeX.py", line 435, in parse
    for item in tokens:
[...]

PR #119, opened 2019-08-27
Small code tidying improvements

llimeht:
In looking for places where `StopIteration` might be an issue (to solve #109), I threw `pylint` at the plastex codebase. Along with the `stop-iteration-return` message that pointed to the right places in the code, `pylint` highlighted lots of other things that were either real problems (such as `collections.abc` imports changing with Python 3.8 and unknown backslash codes becoming errors in a future version of Python 3), while some of pylint's complaints are more stylistic.

To help this task, I created a rudimentary `.pylintrc` that disables most of the warnings that pylint was emitting; I then set about enabling selected warnings and fixing their causes. This staged approach lets the codebase slowly improve and keeps the signal:noise of pylint high such that real errors can be found. Some of pylint's complaints are never worth worrying about and can remain disabled.

Would you be happy to accept a series of patches that systematically works through some of the potential errors and tries to make the code more pythonic in places? This PR is a starting point that fixes a few real bugs and a few cosmetic issues. My experience is that if this process is continued, further bugs will be found by tags such as `undefined-variable` but that it is worth working through some of the easier tags before tackling the harder ones.

(this is based on top of the `fix/generators` branch but can be rebased onto the `python3` branch if you'd prefer)


PatrickMassot:
Merged. Thanks!

PR #121, opened 2019-10-10
Fix escape sequences in strings

llimeht:
Fixes some further backslash sequences found in the unittests.


PatrickMassot:
Thanks for these three PRs @llimeht!

PR #102, opened 2017-10-16
Restore all tests by removing harmful unittests/__init__.py

PatrickMassot:
This reverts 10f51da083851247954aaabc0ad8d540e763c290 from @niklasp which introduced ``unittests/__init__.py``, which messed up imports in tests, which were then marked for skipping.

PR #123, opened 2019-10-10
Add missing classes to setup.py

llimeht:
These classes/resources are otherwise not fully installed or findable by plastex


llimeht:
(Incidentally, the "WARNING: unrecognized command/environment" sorts of errors described in #102 are seen without this patch when trying to test to *installed* version of the plastex module.

PR #132, opened 2019-12-12
Sort index so that output is reproducible

llimeht:
The reproducible builds project is working to make compilation output bit-for-bit reproducible across builds. This makes software independently verifiable and also helps spot accidental changes to code or package contents. 

This patch from Chris Lamb sorts output from the XHTML template and thereby makes the build of the plastex Debian package reproducible.

From https://bugs.debian.org/779595


PatrickMassot:
There is no problem merging this, and I don't think anyone still uses that part of plasTeX. But I'm puzzled by the bug report process. How was the plasTeX team meant to know about this Debian bug report?

llimeht:
The Debian maintainer (as of a couple of months ago, that's me) will forward relevant bug reports and patches to you. Some bugs that the maintainer sees will be entirely to do with packaging and interactions between packages, while others will about the code that lives in this git repository. While we can end up with Debian-specific patches living only within the Debian package, it's generally best if we don't do that.

This one took a while to find its way here partly because plastex had no Debian maintainer for a while. In looking at plastex+python 3 things, I incorporated that patch into the Debian package and verified that it did indeed work to make the 2.0 package reproducible before forwarding it.

PR #133, opened 2019-12-12
Use context manage for chdir in tests

llimeht:
Moving the state around for the tests is error prone. The pytest tmpdir fixture provides a context manager which simplifies these operations greatly.


PatrickMassot:
Thanks!

PR #135, opened 2019-12-12
Fix tikz functional test

llimeht:
The current functional test of the tikz handler needs a few fixups:

* it does not check that `plastex` ran correctly
* it requires additional tools to be installed
* it cannot be run in place (only after installation)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,16 @@ python:
   - "3.7"
   - "3.8"
 
+addons:
+  apt:
+    packages:
+    - dvipng
+    - pdf2svg
+    - texlive-binaries
+    - texlive-latex-base
+    - texlive-latex-extra
+
+
 install:
   - pip install tox-travis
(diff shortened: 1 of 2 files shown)


PatrickMassot:
I don't really understand what you are achieving here. How do you test that the LaTeX packages you install did anything?

llimeht:
The test fails without those packages being installed. From https://travis-ci.org/llimeht/plastex/jobs/624911446
```
...
/bin/sh: 1: dvipng: not found
/bin/sh: 1: dvi2bitmap: not found
/bin/sh: 1: gs: not found
/bin/sh: 1: gs: not found
/bin/sh: 1: cgpdfpng: not found
...
FileNotFoundError: [Errno 2] No such file or directory: 'pdflatex'
```
and then `plastex` exits 1.

Note that in the second of the three commits,  `call()` is changed to `check_call()` so that if running `plastex` fails then the test fails properly.

PR #148, opened 2020-03-23
Try debugging imaging on Travis

PatrickMassot:


PatrickMassot:
This was only for Travis purposes.

PR #149, opened 2020-03-28
HTML5 improvements

dalcde:
1. Add missing `endif` in `Bibliography.jinja2s`
2. Add support for `\textsuperscript` and `\textsubscript`.
3. Add support for `\texorpdfstring`

### DIFF
--- a/plasTeX/Packages/hyperref.py
+++ b/plasTeX/Packages/hyperref.py
@@ -12,6 +12,7 @@ TO DO:
 
 from plasTeX import Command, Environment
 from plasTeX.Base.LaTeX.Crossref import ref, pageref
+from plasTeX.Tokenizer import Tokenizer, Token
 import urllib.parse
 
 def addBaseURL(self, urlarg):
@@ -107,7 +108,7 @@ class pdfstringdef(Command):
     args = 'macroname:str tex:str'
 
 class texorpdfstring(Command):
-    args = 'tex:str pdf:str'
+    args = 'tex pdf:str'
 
 class pdfstringdefDisableCommands(Command):
     args = 'tex:str'
--- a/plasTeX/Packages/hyperref.py
+++ b/plasTeX/Packages/hyperref.py
@@ -12,7 +12,6 @@ TO DO:
 
 from plasTeX import Command, Environment
 from plasTeX.Base.LaTeX.Crossref import ref, pageref
-from plasTeX.Tokenizer import Tokenizer, Token
 import urllib.parse
 
 def addBaseURL(self, urlarg):
--- a/plasTeX/Renderers/HTML5/Bibliography.jinja2s
+++ b/plasTeX/Renderers/HTML5/Bibliography.jinja2s
@@ -21,6 +21,7 @@ name: cite
   <a href="{{ item.idref.bibitem.url }}">{{ item }}</a>
   {% else %}
   {{ item }}
+  {% endif %}
 {% endfor %}
 </span>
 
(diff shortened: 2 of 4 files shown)


dalcde:
I rewrote the `\texorpdfstring` implementation (and force-pushed). I think this is the correct way to do it. The previous way only worked in very specific circumstance.

dalcde:
Edit: this seems to have some issues. I'll work on it a bit more

dalcde:
Should be all good now

PatrickMassot:
Thank you very much for your contribution! It would make things slightly easier if you could separate trivial fixes like the missing `endif` and trivial additions like `textsuperscript` (although that one is misplaced) from the things that require discussion. First I would have merged the trivial fix in 15 seconds. Next we wouldn't need to have two completely orthogonal discussions below.

About `textsuperscript`, because the macro is defined in [Sentences.py](https://github.com/plastex/plastex/blob/master/plasTeX/Base/LaTeX/Sentences.py#L103-L107), it would make more sense to add the templates to [Sentences.jinja2s](https://github.com/plastex/plastex/blob/master/plasTeX/Renderers/HTML5/Sentences.jinja2s), just like in the XHTML renderer (I wonder what happened to those templates when I created the HTML5 renderer from the XHTML one).

About `texorpdfstring` I'm not sure I understand the problem you are trying to solve. And certainly we don't want to throw away half of the arguments. Remember that HTML is not the only plasTeX target, and HTML5+mathjax/KhaTeX is not the only renderer. So maybe some other renderer want to use the string version. What's wrong with having:
```python
class texorpdfstring(Command):
    args = 'tex pdf:str'
```
(note this is not exactly what we currently have) and a HTML5 template that would say:
```
name: texorpdfstring                                         
{{ obj.attributes['tex'] }}
```
If this doesn't work for you then I'd like more examples of TeX sources you want to handle.

dalcde:
>About `textsuperscript`, because the macro is defined in [Sentences.py](https://github.com/plastex/plastex/blob/master/plasTeX/Base/LaTeX/Sentences.py#L103-L107), it would make more sense to add the templates to [Sentences.jinja2s](https://github.com/plastex/plastex/blob/master/plasTeX/Renderers/HTML5/Sentences.jinja2s), just like in the XHTML renderer (I wonder what happened to those templates when I created the HTML5 renderer from the XHTML one).

Fixed.

>About `texorpdfstring` I'm not sure I understand the problem you are trying to solve. And certainly we don't want to throw away half of the arguments. Remember that HTML is not the only plasTeX target, and HTML5+mathjax/KhaTeX is not the only renderer. So maybe some other renderer want to use the string version. What's wrong with having:
>```python
>class texorpdfstring(Command):
>    args = 'tex pdf:str'
>```
>(note this is not exactly what we currently have) and a HTML5 template that would say:
>```
>name: texorpdfstring
>{{ obj.attributes['tex']ï¿½}}
>```
>If this doesn't work for you then I'd like more examples of TeX sources you want to handle.

I swear this didn't work last time I tried, but it seems to be working
fine now... I'll use this instead until I find some issues in the future

plasTeX/Packages/hyperref.py
```diff
@@ -12,6 +12,7 @@
 
 from plasTeX import Command, Environment
 from plasTeX.Base.LaTeX.Crossref import ref, pageref
+from plasTeX.Tokenizer import Tokenizer, Token
```
PatrickMassot: This line is no longer necessary.

PR #150, opened 2020-03-29
Add pdf2svg imager

dalcde:
Currently this uses pdfcrop to crop the pdf before converting it to svg.
I've been unable to find a straightforward way to crop svg's; something
we should probably look into in the future (the Image.crop function
doesn't work on svg's. We should fix that).

The resulting svg's are much smaller than that of dvisvgm, but this is configurable via the scale factor

### DIFF
--- a/plasTeX/Imagers/__init__.py
+++ b/plasTeX/Imagers/__init__.py
@@ -425,6 +425,7 @@ class Imager(object):
 
     # Verification command to determine if the imager is available
     verification = ''
+    verifications = []
 
     fileExtension = '.png'
 
@@ -521,9 +522,16 @@ class Imager(object):
         if self.verification:
             proc = os.popen(self.verification)
             proc.read()
-            if not proc.close():
-                return True
-            return False
+            return not proc.close()
+
+        if self.verifications:
+            for command in self.verifications:
+                proc = os.popen(command)
+                proc.read()
+                if proc.close():
+                    return False
+
+            return True
 
         if not self.command.strip():
             return False
--- a/plasTeX/Imagers/__init__.py
+++ b/plasTeX/Imagers/__init__.py
@@ -9,6 +9,7 @@ from plasTeX.Filenames import Filenames
 from collections import OrderedDict
 import subprocess
 import shlex
+from typing import List, Tuple, Optional
 
 log = getLogger()
 depthlog = getLogger('render.images.depth')
@@ -653,7 +654,7 @@ class Imager(object):
 
         return output
 
-    def executeConverter(self, output):
+    def executeConverter(self, output: bytes) -> Tuple[int, Optional[List[str]]]:
         """
         Execute the actual image converter
 
--- a/plasTeX/Imagers/__init__.py
+++ b/plasTeX/Imagers/__init__.py
@@ -424,7 +424,6 @@ class Imager(object):
     compiler = 'latex'
 
     # Verification command to determine if the imager is available
-    verification = ''
     verifications = []
 
     fileExtension = '.png'
@@ -519,11 +518,6 @@ class Imager(object):
 
     def verify(self):
         """ Verify that this commmand works on this machine """
-        if self.verification:
-            proc = os.popen(self.verification)
-            proc.read()
-            return not proc.close()
-
         if self.verifications:
             for command in self.verifications:
                 proc = os.popen(command)
--- a/plasTeX/Imagers/__init__.py
+++ b/plasTeX/Imagers/__init__.py
@@ -638,12 +638,18 @@ class Imager(object):
                         cmd_line))
         output = None
         for ext in ['.dvi','.pdf','.ps']:
-            if os.path.isfile('images'+ext):
-                output = open('images'+ext, 'rb')
+            try:
+                with open('images'+ext, 'rb') as f:
+                    output = open('images'+ext, 'rb').read()
                 break
+            except OSError:
+                # Ignore if file not found
+                pass
 
         # Change back to original working directory
         os.chdir(cwd)
+        if not self.config["images"]["save-file"]:
+            shutil.rmtree(tempdir, True)
 
         return output
 
@@ -652,7 +658,8 @@ class Imager(object):
         Execute the actual image converter
 
         Arguments:
-        output -- file object pointing to the rendered LaTeX output
+        output -- the content of the rendered LaTeX output, obtained by
+        open(...).read()
 
         Returns:
         two-element tuple.  The first element is the return code of the
@@ -661,7 +668,7 @@ class Imager(object):
         used, you can simply return None.
 
         """
-        open('images.out', 'wb').write(output.read())
+        open('images.out', 'wb').write(output)
         options = ''
         if self._configOptions:
             for opt, value in self._configOptions:
--- a/plasTeX/Imagers/pdf2svg.py
+++ b/plasTeX/Imagers/pdf2svg.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python
+
+import os
+import subprocess
+from plasTeX.Imagers import VectorImager
+
+class PDFSVG(VectorImager):
+    """ Imager that uses pdf2svg """
+    fileExtension = '.svg'
+    verifications = ['pdflatex --help', 'which pdf2svg', 'pdfcrop --help']
+    compiler = 'pdflatex'
+
+    def executeConverter(self, output):
+        open('images-uncropped.pdf', 'wb').write(output.read())
+        subprocess.call(["pdfcrop", "images-uncropped.pdf", "images.pdf"], stdout=subprocess.PIPE)
+
+        rc = 0
+        page = 1
+        while 1:
+            filename = 'img%d.svg' % page
+            rc = subprocess.call(['pdf2svg', 'images.pdf', filename, str(page)], stdout=subprocess.PIPE)
+            if rc:
+                break
+
+            if not open(filename).read().strip():
+                os.remove(filename)
+                break
+            page += 1
+            if page > len(self.images):
+                break
+        return rc, None
+
+Imager = PDFSVG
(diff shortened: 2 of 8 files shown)


plasTeX/Imagers/__init__.py
```diff
@@ -518,12 +518,14 @@ def writePreamble(self, document):
 
     def verify(self):
         """ Verify that this commmand works on this machine """
-        if self.verification:
-            proc = os.popen(self.verification)
-            proc.read()
-            if not proc.close():
-                return True
-            return False
+        if self.verifications:
```
PatrickMassot: This is not directly related to this PR, but since you changed something here, maybe it would be a good opportunity to include more feed-back when this fails. For instance, instead of returning `False` you could return the verification that failed, and then in [`Renderer.__init__`](https://github.com/plastex/plastex/blob/acaa91fbaee4bd826a941d3a8fce18e9a4a18415/plasTeX/Renderers/__init__.py#L450-L452) change the test and write what failed when it fails. If you choose to do that then don't forget to update [the docs](https://github.com/plastex/plastex/blob/8ebba0fa05666da8b0928ea2a40f389f6a64e454/Doc/imager-api.tex#L179-L183).

plasTeX/Imagers/__init__.py
```diff
@@ -636,12 +638,17 @@ def compileLatex(self, source):
                         cmd_line))
         output = None
         for ext in ['.dvi','.pdf','.ps']:
-            if os.path.isfile('images'+ext):
-                output = open('images'+ext, 'rb')
+            try:
+                with open('images'+ext, 'rb') as f:
+                    output = open('images'+ext, 'rb').read()
```
PatrickMassot: This is a pretty risky change. Did you triple checked you propagated it everywhere?

plasTeX/Imagers/__init__.py
```diff
@@ -636,12 +638,17 @@ def compileLatex(self, source):
...
                 break
+            except OSError:
+                # Ignore if file not found
+                pass
 
         # Change back to original working directory
         os.chdir(cwd)
+        shutil.rmtree(tempdir, True)
```
PatrickMassot: This is yet another unrelated change, right? It is often useful to keep those temporary stuff for debugging, and I don't think they are so large (and people are still free to wipe out `/tmp` from times to times).

plasTeX/Imagers/pdf2svg.py
```diff
@@ -0,0 +1,56 @@
...
+
+import subprocess
+import re
+import shutil
+from plasTeX.Imagers import VectorImager
+
+width_re = re.compile('width="([0-9\\.]*)')
+height_re = re.compile('height="([0-9\\.]*)')
+class PDFSVG(VectorImager):
+    """ Imager that uses dvisvgm """
```
PatrickMassot: Copy-paste error?

plasTeX/Imagers/pdf2svg.py
```diff
@@ -0,0 +1,56 @@
...
+                break
+    
+            scale = self.config["images"]["scale-factor"]
+            if scale == 1:
+                continue
+
+            scaled = False
+            with open("scaled-%s" % filename, "w") as ofile:
+                for line in open(filename, "r"):
+                    if not scaled and "<svg" in line:
```
PatrickMassot: All this SVG parsing seems brittle (see also the first answer on [this page](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags)). In the current TikZ handling we use BeautifulSoup. Maybe this is a bit extreme, but there is xml parsing in the python standard lib.

plasTeX/Imagers/__init__.py
dalcde: Yeah. I grep'ed the source and made sure it is only propagated to `executeConvert`, and I modified all the `executeConvert`'s.

plasTeX/Imagers/pdf2svg.py
dalcde: I agree. My original rationale was that we only had to parse very specific output, namely what pdf2svg gives us, but I suppose this doesn't excuse bad code. I'll fix this.

plasTeX/Imagers/__init__.py
dalcde: I also use this for debugging sometimes, but it is less helpful if the `/tmp/` is full of things called `tmp....`. Personally, I use this on a shared host, and it would be irresponsible to pollute `/tmp/` with all my stuff. I'll add an option that removes the temp directories by default.

dalcde:
All done

plasTeX/Imagers/__init__.py
dalcde: I suppose one way to do this is for `verify` to raise an exception when it is missing something, and the `Renderer` would catch this error and print the corresponding message.

PatrickMassot: Yes, let's keep that for a future PR.

plasTeX/Imagers/pdf2svg.py
```diff
@@ -0,0 +1,46 @@
...
+
+length_re = re.compile('([0-9\\.]*)(.*)')
+class PDFSVG(VectorImager):
+    """ Imager that uses pdf2svg """
+    fileExtension = '.svg'
+    verifications = ['pdflatex --help', 'which pdf2svg', 'pdfcrop --help']
+    compiler = 'pdflatex'
+
+    def executeConverter(self, output):
+        open('images-uncropped.pdf', 'wb').write(output)
```
PatrickMassot: I know plasTeX is full of code like this. But it was written 15 years ago, let's not add more. Please use 
```python
with open('images-uncropped.pdf', 'wb') as uncropped:
    uncropped.write(output)
```
which makes sure to close the file handler immediately, or even more modern: start the file with `from pathlib import Path` and then use `Path('images-uncropped.pdf').write_bytes(output)` instead of the above line, since the next plasTeX release will support only python at least 3.5.

plasTeX/Imagers/pdf2svg.py
```diff
@@ -0,0 +1,46 @@
...
+        subprocess.call(["pdfcrop", "images-uncropped.pdf", "images.pdf"], stdout=subprocess.DEVNULL)
+
+        rc = 0
+        page = 1
+        while 1:
```
PatrickMassot: I don't understand why this is a `while` loop. Isn't it iterating over `self.images`? Or at least `range(1, len(self.images)+1)` since you seem to use only the page number.

plasTeX/Imagers/pdf2svg.py
dalcde: Glad to hear about the last part. Does that mean I'm allowed to add type hints as I go?

plasTeX/Imagers/pdf2svg.py
dalcde: Yeah I think we should do that instead... I really shouldn't blindly copy existing code.

dalcde:
On Tue, Mar 31, 2020 at 03:35:28AM -0700, Patrick Massot wrote:
>It would be nice to add a test case. Please ask if you don't know how to write tests here.

I've looked into the testing but couldn't figure out how it works. Some
pointers would be helpful.

plasTeX/Imagers/pdf2svg.py
PatrickMassot: Yes! You can add as many type hints as you wish, our lives would be much easier with type hints in plasTeX. Even without using mypy, it's already great documentation. But let's keep it python 3.5 compatible for now (so no PEP 526 variable type hints).

plasTeX/Imagers/__init__.py
```diff
@@ -638,12 +638,18 @@ def compileLatex(self, source):
...
+                with open('images'+ext, 'rb') as f:
+                    output = open('images'+ext, 'rb').read()
                 break
+            except OSError:
+                # Ignore if file not found
+                pass
 
         # Change back to original working directory
         os.chdir(cwd)
+        if not self.config["images"]["save-file"]:
```
PatrickMassot: Now I'm confused. I was about to write this should be added in https://github.com/plastex/plastex/blob/master/plasTeX/Config.py but I see it's already there, added 14 years ago in https://github.com/plastex/plastex/commit/d9c53162a3252d0d5651f0d3fe16d4ef76dc5ab0. Did you see it used anywhere? It would be nice to dig a bit into the git history of the project to see what happened here. In any case, did you make sure this option is mentioned in the docs?

dalcde:
On Tue, Mar 31, 2020 at 07:53:46AM -0700, Patrick Massot wrote:
>
>
>
>>
>         # Change back to original working directory
>         os.chdir(cwd)
>+        if not self.config["images"]["save-file"]:
>
>Now I'm confused. I was about to write this should be added in https://github.com/plastex/plastex/blob/master/plasTeX/Config.py but I see it's already there, added 14 years ago in https://github.com/plastex/plastex/commit/d9c53162a3252d0d5651f0d3fe16d4ef76dc5ab0.

I grep'ed the whole codebase and it isn't used anywhere. The option
shows up when you run `plastex --help`, but it isn't in the docs proper.
I've just added it.

plasTeX/Imagers/__init__.py
PatrickMassot: Ok, good.

plasTeX/Renderers/__init__.py
```diff
@@ -472,6 +472,8 @@ def render(self, document, postProcess=None):
                 break
             elif name == 'dvisvgm':
                 from plasTeX.Imagers.dvisvgm import Imager as VectorImager
+            elif name == 'pdf2svg':
+                from plasTeX.Imagers.pdf2svg import Imager as VectorImager
             else:
                 log.warning("Invalid imager '%s'")
```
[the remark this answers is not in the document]
PatrickMassot: Forget about this, I'll do it.

PatrickMassot:
I decided it would be too complicated to tell you how to write a test for this. So I'll merge, add a test, and discuss later. I did add a minimal [CONTRIBUTE.md](https://github.com/plastex/plastex/blob/master/CONTRIBUTING.md) that you can already read.

PatrickMassot:
Thanks again!

PatrickMassot:
@dalcde please see d40b7fbb7b40a577961205d036338f223e5fb4bf about testing an imager.

PR #151, opened 2020-04-04
Use Imager for Tikz

dalcde:
This moves verbatim to a `VerbatimEnvironment` in `plasTeX/__init__.py` and
simplifies some of the code, so that we can use it for classes we intend
to use imager for.

This also rewrites `NoCharSubEnvironment` using what `verbatim` does. This
is cleaner because it overwrites `normalize` instead of `invoke`, and more
classes will want to have custom `invoke` than custom `normalize`. This
means it is generally easier to inherit `NoCharSubEnvironment`.

I haven't updated the tests yet, but want to put this out for discussion first.


dalcde:
RE the config options removed:
 - I think the template configuration is unnecessary, because the `imager` reuses the full preamble of the original tex file.
 - The `tikz-converter` is superceded by the `imager` choice.
 - The `tikz-compiler` option is still potentially useful. Currently this choice is bundled with the choice of the imager, since not all imagers support all compilers (e.g. `pdf2svg` must use `pdflatex` instead of `latex`, since `latex` produces a `dvi`). I think it should be up to the imager to make this configurable.

PatrickMassot:
About the compiler option, I think we need a broader change. We need a global option saying what is our favorite TeX compiler, and then each imager can declare a list of supported compilers, or at least of supported input format, so that we can raise a warning and not use an imager if it does not support this desired compiler or output format. 

The discussion is not longer between latex and pdflatex. I don't see how anyone could want to produce a dvi file in 2020. The discussion is between pdfLaTeX, XeLaTeX and LuaLaTeX, all producing pdf. In particular the use of `unicode-math` is crucial. plasTeX has no issue compiling documents containing unicode in math, but `pdflatex` cannot be used here.

I think this should be done in a separate PR, while that one is waiting. I really appreciate your work, but you're doing too much at the same time. Also I don't understand why you destroyed my imager tests while you could have added a new test file. My tests were faster and less sensitive to irrelevant changes. In your version you are testing the HTML5 renderer along with the imager. This is no longer a unit test, and it means any change in the renderer will need to change at the least the benchmark in your test.

So I cannot merge this PR, but you could:
* Open a PR containing only the changes to verbatim. We already have unit tests for verbatim, so hopefully they still work and that PR would be merged very fast.
* Open a PR about specifying a compiler as discussed above, together with tests ensuring this works (preferably without testing the HTML5 renderer at the same time).
* Once the above two PRs are merged, a PR introducing the tikz switch, with tests (without changing the existing simple imager tests).

dalcde:
On Mon, Apr 06, 2020 at 04:26:22AM -0700, Patrick Massot wrote:
>About the compiler option, I think we need a broader change. We need a global option saying what is our favorite TeX compiler, and then each imager can declare a list of supported compilers, or at least of supported input format, so that we can raise a warning and not use an imager if it does not support this desired compiler or output format.
>
>The discussion is not longer between latex and pdflatex. I don't see how anyone could want to produce a dvi file in 2020. The discussion is between pdfLaTeX, XeLaTeX and LuaLaTeX, all producing pdf. In particular the use of `unicode-math` is crucial. plasTeX has no issue compiling documents containing unicode in math, but `pdflatex` cannot be used here.

What is the plan with imagers such as dvisvgm and dvipng then? Do we
drop support for them?

>So I cannot merge this PR, but you could:
>* Open a PR containing only the changes to verbatim. We already have unit tests for verbatim, so hopefully they still work and that PR would be merged very fast.
>* Open a PR about specifying a compiler as discussed above, together with tests ensuring this works (preferably without testing the HTML5 renderer at the same time).
>* Once the above two PRs are merged, a PR introducing the tikz switch, with tests (without changing the existing simple imager tests).

Will do

PatrickMassot:
> On Mon, Apr 06, 2020 at 04:26:22AM -0700, Patrick Massot wrote: About the compiler option, I think we need a broader change. We need a global option saying what is our favorite TeX compiler, and then each imager can declare a list of supported compilers, or at least of supported input format, so that we can raise a warning and not use an imager if it does not support this desired compiler or output format. The discussion is not longer between latex and pdflatex. I don't see how anyone could want to produce a dvi file in 2020. The discussion is between pdfLaTeX, XeLaTeX and LuaLaTeX, all producing pdf. In particular the use of `unicode-math` is crucial. plasTeX has no issue compiling documents containing unicode in math, but `pdflatex` cannot be used here.
> What is the plan with imagers such as dvisvgm and dvipng then? Do we drop support for them?
No, they will declare they support only dvi files, so every time plasTeX will be run with a favorite compiler option implying pdf, those imagers won't be tried. This is the same thing as with failing verifications in the existing code.

dalcde:
Makes sense

dalcde:
I have rebased the commits on top of the current master. The new test no longer replaces the old imager test and is at `unittests/tikz-tikzcd.py`.

I believe the unit test should not only test that the image is correctly produced, but also that it is correctly rendered into the resulting HTML (this necessarily involves the choice of rendering mechanism, because other renderers may opt to not use the imager). On the other hand, this can be made more robust by extracting the tags that say `<div class="tikzpicture">` or `<div class="tikzcd">`.

PatrickMassot:
The tests are still failing. Are you working on this?

PatrickMassot:
It was partly my mistake, my python 3.5 hack was not good enough see 76e1051. Supporting python 3.5 is really a chore.

PatrickMassot:
Oh crap, you also fixed it. I hope my fix is better.

dalcde:
The current status is that the tests work on my machine but not travis.

Part of this is because I have a different version of the relevant
pdf-to-png converter, which produces slightly different output.

I have no idea why the html output is different. They differ in the id
attribute which should be computed by plasTeX.

The imager tests are also still failing

PatrickMassot:
I think the mysterious id mismatch is because you use `render(tex.parse())` instead of `tex.parse()` followed by `render(doc)` (but if you say pytest works on your end then I'm confused).

The tiny png mismatch is worse worse, and I hate those things. Maybe try to find an imager which is stable enough (ie. abandoned) to have the same version everywhere. If this doesn't work then let's drop comparing the bitmap image. We don't want to encourage people to use them rather than svg anyway...

dalcde:
The test is now passing. This uses imagemagick to compare the resulting images and passes if the difference is within an acceptable threshold. This is not fool proof but definitely better than checking if the files are bytewise equal. I'm happy to port this to the imager tests if you wish.

PatrickMassot:
FYI, I started on fixing the SVG comparison in https://github.com/plastex/plastex/commits/travis-images. In particular Travis now shows the differences: https://travis-ci.org/github/plastex/plastex/jobs/676703102

But this is pretty discouraging, with diffs like:
```
-  <use xlink:href="#glyph0-1" x="-0.056" y="7.875"/>
+  <use xlink:href="#glyph0-1" x="-0.056" y="7.765"/>
```
And of course things like:
```
- <!-- This file was generated by dvisvgm 2.6.3 -->
+ <!-- This file was generated by dvisvgm 1.9.2 -->
```
If you don't have a better idea I'm afraid we'll have to convert everything into png before resorting to the imagemagick compare stuff.

dalcde:
My impression was that imagemagick can work directly with svg's as well.
On the other hand, being able to see the difference in svg is helpful.

A minor improvement to comparing the texts would be to parse the svgs
and compare the trees for equality.

PatrickMassot:
Indeed I was able to use it for svg (of course it does convert to a bitmap format under the hood). I also bumped the Ubuntu version used by Travis and this got rid of most discrepancies. But dvipng stays outdated and I mostly gave up and relaxed the test. I already spend way too much time trying to debug this, and we really need tests that work for all the work ahead.

dalcde:
When ImageMagick 7 finally lands in Ubuntu, we can switch to the DSSIM
metric which seems to work much better. Sadly, focal still has 6.9.10 so
we'll have to wait two years...

PatrickMassot:
We could locally install version 7 at each build, but let's keep that as a last resort.

PR #154, opened 2020-04-06
Move verbatim to VerbatimEnvironment in __init__.py

dalcde:
This also rewrites NoCharSubEnvironment using what verbatim does. This
is cleaner because it overwrites normalize instead of invoke, and more
classes will want to have custom invoke than custom normalize. This
means it is generally easier to inherit NoCharSubEnvironment.

### DIFF
--- a/plasTeX/__init__.py
+++ b/plasTeX/__init__.py
@@ -914,27 +914,97 @@ class Environment(Macro):
         if dopars:
             self.paragraphs()
 
-
 class NoCharSubEnvironment(Environment):
     """
     A subclass of Environment which prevents character substitution inside
     itself.
     """
-    def __init__(self, *args, **kwargs):
-        # Will hold the owner document charsubs to restore it at the end
-        self.charsubs = []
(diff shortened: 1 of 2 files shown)


plasTeX/__init__.py
```diff
@@ -914,27 +914,97 @@ def digest(self, tokens):
...
     itself.
     """
-    def __init__(self, *args, **kwargs):
-        # Will hold the owner document charsubs to restore it at the end
-        self.charsubs = []
-        super(NoCharSubEnvironment, self).__init__(*args, **kwargs)
+
+    def normalize(self, charsubs=None):
+        """ Normalize, but don't allow character substitutions """
+        return Environment.normalize(self)
```
PatrickMassot, line 925: I think it would be much clearer to write:
```suggestion
        return Environment.normalize(self, charsubs=None)
```
rather than relying on the fact this is the default value anyway.

dalcde:
Fixed

PatrickMassot:
Thanks!

PR #155, opened 2020-04-06
Allow specifying different compilers for Imager and VectorImager

dalcde:
Clarify documentation and unit test with different compilers.


PatrickMassot:
All this is good but will fail mysteriously if the imager and compiler options are incompatible. I think each imager should explicitly declare its accepted input formats. Then the `images['compiler']` and `images['vector-compiler']` could parse their values as follows:
* if the value is "foo, bar" then (after stripping white space), foo becomes the compiler value, and bar is its output filename extension, such as dvi or pdf.
* else look up the extension in a hard-coded list of (foo, bar) pairs, containing [(latex, dvi), (pdflatex, pdf), etc.].

This way we don't have to specify obvious file extensions but the user can still use an exotic compiler. And each imager can test whether the compiler it would call has compatible format.

dalcde:
This would break backwards compatibility, since the images['compiler']
option already exists and is respected (and has the current behaviour).

One possible option is to have a whitelist of allowed compilers, and
warn when the compiler is not on the list. I would like to still give
users the chance to use some exotic compiler they found in their garage
if they wish to do so.

PatrickMassot:
My proposal also allowed to use a garage compiler. But let's say the user takes responsibility for not asking incompatible imager and compiler (this is a python program after all).

PR #158, opened 2020-04-12
Move compilation logic from cli to library

dalcde:
This makes it easier for people to write their own custom build logic

### DIFF
--- a/plasTeX/Compile.py
+++ b/plasTeX/Compile.py
@@ -60,7 +60,7 @@ def run(filename: str, config: ConfigManager):
         rmodule = importlib.import_module('plasTeX.Renderers.'+rname)
     except ImportError as msg:
         log.error('Could not import renderer "%s".  Make sure that it is installed correctly, and can be imported by Python.' % rname)
-        sys.exit(1)
+        raise msg
 
     # Write expanded source file
     #sourcefile = '%s.source' % jobname
--- a/plasTeX/Compile.py
+++ b/plasTeX/Compile.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python
+
+import os, sys, string, glob
+import importlib
+import plasTeX
+from plasTeX.TeX import TeX
+from plasTeX.ConfigManager import *
+from plasTeX.Logging import getLogger, updateLogLevels
+
+log = getLogger()
+
+def run(filename: str, config: ConfigManager):
+    if config['logging']:
+        updateLogLevels(config['logging'])
+
+    # Create document instance that output will be put into
+    document = plasTeX.TeXDocument(config=config)
+
+    # Instantiate the TeX processor and parse the document
+    tex = TeX(document, myfile=filename)
+
+    # Send log message to file "jobname.log" instead of console
+    if config['files']['log']:
+        tex.fileLogging()
+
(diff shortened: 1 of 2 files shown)


PatrickMassot:
As you know, I think this is a worthwhile goal, but this PR is far from being enough, and I don't even see how it helps. 

The main thing is we need to get rid of 
```python
rmodule = importlib.import_module('plasTeX.Renderers.'+rname)
```
which imposes the renderer is distributed with plasTeX. The minimal modification would be to allow a fully qualified name in the `--renderer` option, and use `plasTeX.Renderers` as a fall back prefix for unqualified names. A better way would be to use some setuptools magic so that python packages can declare they provide plasTeX renderers, as in [pluggy](https://pluggy.readthedocs.io/en/latest/). Note that I'm not suggesting we add pluggy as a dependency, that would be a huge overkill. I only suggest we take inspiration. It's really convenient that, for instance, pytest plugins can modify the behavior of pytest simply by being installed, without touching pytest code.

The second point is we could have `Compiler` class having a `run` method calling a bunch of other methods like: `load_config`, `load_aux`, `prepare_outdir`, `load_renderer` etc. Then people who want a custom cli could subclass `Compiler` and override whatever they want without copy-pasting code.

I intended to do all that myself, but this you opened this PR, feel free to contribute (and discuss)!

dalcde:
On Mon, Apr 13, 2020 at 06:07:53AM -0700, Patrick Massot wrote:
>The main thing is we need to get rid of
>```python
>rmodule = importlib.import_module('plasTeX.Renderers.'+rname)
>```
>which imposes the renderer is distributed with plasTeX. The minimal modification would be to allow a fully qualified name in the `--renderer` option, and use `plasTeX.Renderers` as a fall back prefix for unqualified names. A better way would be to use some setuptools magic so that python packages can declare they provide plasTeX renderers, as in [pluggy](https://pluggy.readthedocs.io/en/latest/). Note that I'm not suggesting we add pluggy as a dependency, that would be a huge overkill. I only suggest we take inspiration. It's really convenient that, for instance, pytest plugins can modify the behavior of pytest simply by being installed, without touching pytest code.
>
>The second point is we could have `Compiler` class having a `run` method calling a bunch of other methods like: `load_config`, `load_aux`, `prepare_outdir`, `load_renderer` etc. Then people who want a custom cli could subclass `Compiler` and override whatever they want without copy-pasting code.

I agree with all of this and was intending to do some of it. However,
any such change would necessarily conflict with the current PR, so I
wanted to get this (hopefully uncontroversial) change merged before
making bigger changes.

PatrickMassot:
I still don't understand the point of this PR. What does it allow?

dalcde:
It makes the existing code more reusable since it is now part of the
library instead of the binary. This would be a necessary step if we want
to make a reusable Compiler class.

That said, with this PR merged, if I have some custom build script, I
can directly run the build command with a configuration object instead
of turning the configuration into command line arguments and then back
to a Config object.

PatrickMassot:
Ok, let's merge this and move forward.

PatrickMassot:
@dalcde I just created https://github.com/plastex/plastex/milestone/1 to coordinate ongoing efforts.

PR #166, opened 2020-04-15
Allow using Renderer given by a path

### LINKED ISSUE #159: Make it easier to use custom rendering
This issue tracks efforts to make it easier to use custom rendering, including custom configuration handling and using a renderer which does not ship with plasTeX. The following is copy-pasted from the discussion at #158 

The main thing is we need to get rid of 
```python
rmodule = importlib.import_module('plasTeX.Renderers.'+rname)
```
which imposes the renderer is distributed with plasTeX. The minimal modification would be to allow a fully qualified name in the `--renderer` option, and use `plasTeX.Renderers` as a fall back prefix for unqualified names. A better way would be to use some setuptools magic so that python packages can declare they provide plasTeX renderers, as in [pluggy](https://pluggy.readthedocs.io/en/latest/). Note that I'm not suggesting we add pluggy as a dependency, that would be a huge overkill. I only suggest we take inspiration. It's really convenient that, for instance, pytest plugins can modify the behavior of pytest simply by being installed, without touching pytest code.

The second point is we could have `Compiler` class having a `run` method calling a bunch of other methods like: `load_config`, `load_aux`, `prepare_outdir`, `load_renderer` etc. Then people who want a custom cli could subclass `Compiler` and override whatever they want without copy-pasting code.

dalcde:
Working towards #159

### DIFF
--- a/Doc/command.tex
+++ b/Doc/command.tex
@@ -77,7 +77,8 @@ files and packages.
 \options{\longprogramopt{renderer=\optval{renderer-name}}}
 \config{general}{renderer}
 \default{XHTML}
-specifies which renderer to use.
+specifies which renderer to use. This is either one of the built in renderers,
+or a path to the directory of a renderer.
 \end{configuration}
 
 \begin{configuration}{Themes}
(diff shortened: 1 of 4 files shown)


PatrickMassot:
Thanks! What about writing a test?

Issue #160, opened 2020-04-14
Allow Gerby to use upstream plasTeX


PatrickMassot:
The [stacks project](https://stacks.math.columbia.edu/) uses plasTeX for rendering with a [very unusual workflow](https://gerby-project.github.io/). They currently use a [plasTeX fork](https://github.com/gerby-project/plastex) whereas they should be able to use upstream plasTeX with a custom renderer and cli. This is issue is a specific milestone for #159

dalcde:
This is a list of all gerby commits as of 2020-04-20, minus some I know have been upstreamed/are irrelevant. A commit is considered irrelevant if it only affects things in the Gerby renderer or the HTML5 renderer (which afaik Gerby doesn't use). There is I believe a large number of commits here that are irrelevant in this sense, but I haven't gone through all of them

 - [ ] Started linking labels and tags in the DOM tree 7033c96
 - [ ] Detects tags at Renderable.__str__ level 9b06d06
 - [ ] Copying and modifying the bare essentials of HTML5 639c295
 - [ ] Started a Gerby configuration file de294e9
 - [ ] Use plasTeX's output routine to generate files 39c8750
 - [ ] Implemented the marginnote package 1339535
 - [ ] Fixed filename generator 947783d
 - [ ] Partial statement finder for proofs b76327e
 - [ ] Made searchTheorem a static method 992f502
 - [ ] Implemented output propagation 7f7fa2b
 - [ ] Moved tags to userdata b696355
 - [ ] No file extensions for proofs b169a6f
 - [ ] Code cleanup and \ref in proof caption matching 5a3ec72
 - [ ] Make \cite handling more stupid abc268d
 - [ ] Adding slogan environment 103b7f4
 - [ ] Less clutter files by setting split-level option 2c87cf8
 - [ ] Improved templates bb3f774
 - [ ] Using file extensions for output 5576ddd
 - [ ] Handling slogans 3a87e34
 - [ ] Implement history environment 3d43dfb
 - [ ] Fixed silly template issue 4b85205
 - [ ] Small output changes aa226d2
 - [ ] Reordered some lines a9c66f6
 - [ ] Changed theorem template 42caca0
 - [ ] Temporarily fixing Jacob's multi-file indexing 1e0f757
 - [ ] Fixes #2 67f98b3
 - [ ] Fixes #5 0f1f235
 - [ ] Fixes #3 992df1b
 - [ ] Ability to add custom markers and change default style to match LaTeX.  717ba0e
[...]

PR #168, opened 2020-04-15
Backport some fixes from gerby-project

dalcde:
The first commit fixes spacing issues when the theorem content only has one paragraph.

The second commit marks `proof` as `blockType`.

### DIFF
--- a/plasTeX/Base/LaTeX/Definitions.py
+++ b/plasTeX/Base/LaTeX/Definitions.py
@@ -86,11 +86,11 @@ class newtheorem(Command):
         if attrs['*modifier*']:
             newclass = type(str(name), (Environment,),
                     {'caption': caption, 'nodeName': 'thmenv', 'thmName': name,
-                        'args': '[title]'})
+                        'args': '[title]', 'forcePars': True})
         else:
             newclass = type(str(name), (Environment,),
                     {'caption': caption, 'nodeName': 'thmenv', 'thmName': name,
-                        'counter': counter, 'args': '[title]'})
+                        'counter': counter, 'args': '[title]', 'forcePars': True})
         self.ownerDocument.context.addGlobal(name, newclass)
 
 
--- a/plasTeX/Base/LaTeX/Definitions.py
+++ b/plasTeX/Base/LaTeX/Definitions.py
@@ -95,6 +95,7 @@ class newtheorem(Command):
 
 
 class proof(Environment):
+    blockType = True
     args ='[caption]'
 
     def digest(self, tokens):
--- a/unittests/Proof.py
+++ b/unittests/Proof.py
@@ -1,0 +1,48 @@
+from plasTeX.TeX import TeX
+
+def test_proof_in_par():
+    """Check proof environment ends the current paragraph."""
+    tex = TeX()
+    tex.input(r'''
+      \documentclass{article}
+      \begin{document}
+
+      a
+      \begin{proof}
+      b
+      \end{proof}
+      c
+
+      \end{document}
+      ''')
+    doc = tex.parse()
+    proof = doc.getElementsByTagName('proof')[0]
+    if len(proof.parentNode.childNodes) != 1:
+        print(doc.toXML())
+        assert False
+
+def test_proof_block_type():
+    """
+    Check proof environment ends up inside a blockType paragraph. This in
+    particular makes HTML5 not wrap the resulting object in <p></p> tags.
+    """
+    tex = TeX()
+    tex.input(r'''
+      \documentclass{article}
+      \begin{document}
+
+      a
+
+      \begin{proof}
+      b
+      \end{proof}
+
+      c
+
+      \end{document}
+      ''')
+    doc = tex.parse()
+    proof = doc.getElementsByTagName('proof')[0]
+    if not proof.parentNode.blockType:
+        print(doc.toXML())
+        assert False
--- a/unittests/Theorem.py
+++ b/unittests/Theorem.py
@@ -1,0 +1,15 @@
+from plasTeX.TeX import TeX
+from plasTeX.Base.TeX.Primitives import par
+
+def test_single_paragraph_theorem():
+    """Check single paragraph theorem does contain a paragraph"""
+    tex = TeX()
+    tex.input(r'''
+      \newtheorem{thm}{Theorem}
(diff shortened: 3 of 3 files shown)


PatrickMassot:
I guess there is no problem, but let's ping @chngr and @pbelmans to make sure they are happy you are helping with #160.

It would be nice to have a test clearly demonstrating what is fixed.

pbelmans:
Yes, I would be more than happy to help, and thanks for the amazing work you've already done! I've been following the developments a bit, but I haven't had time yet to actually contribute.

dalcde:
On Thu, Apr 16, 2020 at 02:18:46AM -0700, Patrick Massot wrote:
>I guess there is no problem, but let's ping @chngr and @pbelmans to make sure they are happy you are helping with #160.

At the moment I'm merely backporting changes that I personally use. I
can potentially help with #160 but it is not the intention of the
current PR. I try my best to cherry-pick commits from gerby-project to
give the right people credit, but sometimes it is difficult.

>It would be nice to have a test clearly demonstrating what is fixed.

I'm not quite sure what tests I can write.

For the first one, the theorem environment puts the content of the
theorem in a paragraph (<p> tag in html) iff there is more than one
paragraph. This leads to inconsistent spacing, because <p></p> generally
puts some surrounding spaces.

For the second one, a blockType environment automatically ends previous
paragraphs if present. Currently, if we have
```tex
a
\begin{proof}
b
\end{proof}
c
```
we end up with
```html
<p>a <div class="proof_wrapper"> ... </div> c</p>
```
By making `proof` blockType, we would instead have
```html
<p>a</p> <div class="proof_wrapper"> ... </div> <p>c</p>
```
Technically, the former is invalid html, since paragraphs cannot
supposed to include "block" elements. However, most html parsers are
pretty lenient and parses this without much issues (albeit possibly with
weird spacing).

In general, the issue with these is that they produce slightly less
optimal html (and presumably for other renderers too). This would be
difficult to test short of asking for a specific generated html/parsed
tree, which would be pretty fragile (I learnt my lesson with the tikz
stuff)

dalcde:
>I guess there is no problem, but let's ping @chngr and @pbelmans to make sure they are happy you are helping with #160.

At the moment I'm merely backporting changes that I personally use. I
can potentially help with #160 but it is not the intention of the
current PR. I try my best to cherry-pick commits from gerby-project to
give the right people credit, but sometimes it is difficult.

>It would be nice to have a test clearly demonstrating what is fixed.

I'm not quite sure what tests I can write.

For the first one, the theorem environment puts the content of the
theorem in a paragraph (<p> tag in html) iff there is more than one
paragraph. This leads to inconsistent spacing, because <p></p> generally
puts some surrounding spaces.

For the second one, a blockType environment automatically ends previous
paragraphs if present. Currently, if we have
```tex
a
\begin{proof}
b
\end{proof}
c
```
we end up with
```html
<p>a <div class="proof_wrapper"> ... </div> c</p>
```
By making `proof` blockType, we would instead have
```html
<p>a</p> <div class="proof_wrapper"> ... </div> <p>c</p>
```
Technically, the former is invalid html, since paragraphs cannot
supposed to include "block" elements. However, most html parsers are
pretty lenient and parses this without much issues (albeit possibly with
weird spacing).

In general, the issue with these is that they produce slightly less
optimal html (and presumably for other renderers too). This would be
difficult to test short of asking for a specific generated html/parsed
tree, which would be pretty fragile (I learnt my lesson with the tikz
stuff)

pbelmans:
> I try my best to cherry-pick commits from gerby-project to give the right people credit, but sometimes it is difficult.

Don't worry about cherry-picking commits from Gerby from me!

chngr:
This looks great! Glad some of our work is helpful to plasTeX! Feel free to cherry pick commits from myself, as well!

PatrickMassot:
Ok, I tried to write tests. First please add a file `unittests/theorem.py` containing:
```python
from plasTeX.TeX import TeX
from plasTeX.Base.TeX.Primitives import par

def test_single_paragraph_theorem():
    """Check single paragraph theorem does contain a paragraph"""
    tex = TeX()
    tex.input(r'''
      \newtheorem{thm}{Theorem}
      \begin{thm}
      $1+1=2$
      \end{thm}
      ''')
    nodes = tex.parse().getElementsByTagName('thm')[0].childNodes
    assert len(nodes) == 1
    assert isinstance(nodes[0], par)
```
That one works nicely. It fails before your patch, and works after.

For proofs I haven't been able to observe a fix. I tried adding `unittests/proof.py` containing:
```python
from plasTeX.TeX import TeX
from plasTeX.Base.TeX.Primitives import par

def test_proof_in_par():
    """Check proof environment ends the current paragraph."""
    tex = TeX()
    tex.input(r'''
      \documentclass{article}
      \begin{document}

      a
      \begin{proof}
      b
      \end{proof}
      c

      \end{document}
      ''')
    doc = tex.parse()
    proof = doc.getElementsByTagName('proof')[0]
    assert not isinstance(proof.parentNode, par)

def test_proof_dont_create_par():
    """Check proof environment don't end up inside a paragraph"""
    tex = TeX()
    tex.input(r'''
      \documentclass{article}
      \begin{document}

      a

      \begin{proof}
      b
      \end{proof}

      c

      \end{document}
      ''')
    doc = tex.parse()
    proof = doc.getElementsByTagName('proof')[0]
    assert not isinstance(proof.parentNode, par)
```
Which fail, but your patch doesn't seem to fix it. Did I misunderstand the issue? (tests also helps communicating *what* is the issue you want to fix).

Is it clearer what I mean by test? Do you know you can [ask pytest to run only selected tests](https://docs.pytest.org/en/latest/usage.html#specifying-tests-selecting-tests)?  For instance here try `pytest -k proof`. Or better, `pytest -k proof --pdb` to drop into pdb when the assertion fails (after installing [pdbpp](https://github.com/pdbpp/pdbpp) of course).

dalcde:
I've added the tests. `plasTeX` puts *everything* inside `par` environments. However, blockType objects are put in their own `par` instance, and the `par` instance is marked `blockType`. The HTML5 renderer checks if the `par` instance is `blockType`, and if its, it omits the `<p></p>` tags.

PatrickMassot:
Great! Tests make everything much clearer, now and in the distant future.

Issue #137, opened 2019-12-31
Shorthand for German umlauts


pbelmans:
Using `\usepackage[german]{babel}` it is possible to write the often-used German diacritics (such ä, or ö) as `"a` instead of `\"a`.

These are currently not supported, with no replacement happening, hence one gets `"a` in the output.

PatrickMassot:
Why don't you write it as ä?

pbelmans:
It's not my own text I'm experimenting on, so I'm not making the calls on how the TeX source is written, and amongst German TeX'ers writing `"a` seems to be quite common.

A search & replace would do the trick, but if implementing it directly in plasTeX is not too hard, it'd make the applicability of plasTeX a bit bigger.

PR #169, opened 2020-04-16
Fix `Macro.source` for non-letter commands

dalcde:
If a command name comprises of a single non-letter, then there should
not be a space between the command and the argument because the command
is necessarily one character long.

One can test this with $\text{\"o}$ using the HTML5 renderer.

This is fix is not entirely ideal. In theory, we should decide if the name is a non-letter during parsing. However, `EscapeSequence` transitively inherits `str`, and hence is immutable, so we cannot record this information. In general, we really should just remember what the source is during parsing, instead of trying to second guess it based on the parsed result.

### DIFF
--- a/unittests/accents.py
+++ b/unittests/accents.py
@@ -1,0 +1,8 @@
+from plasTeX.TeX import TeX
+
+def test_accent():
+    input_data = r'\"o'
+    tex = TeX()
+    tex.input(input_data)
+    node = tex.parse()[0]
+    assert node.source == input_data
--- a/plasTeX/__init__.py
+++ b/plasTeX/__init__.py
@@ -421,7 +421,8 @@ class Macro(Element):
         argSource = sourceArguments(self)
         if not argSource:
             argSource = ' '
-        elif argSource[0] in encoding.stringletters():
+        elif argSource[0] in encoding.stringletters() and\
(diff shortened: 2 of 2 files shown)


PatrickMassot:
I have no idea what you are talking about. I tried with 
```latex
\documentclass{article}

\usepackage{amsmath}

\begin{document}
	$\text{"o}$
\end{document}
```
and I can't see any difference between the output of `pdflatex` and `plastex --renderer=HTML5  --dollars`.

Again it would help a lot if you could write *tests* each time you propose a fix. Those tests should fail on master and work in the PR of course.

PatrickMassot:
Oh, seeing Pieter's avatar in the other discussion reminded my the existence of #137. Is this the same thing?

dalcde:
I've pushed a test. Without the fix, `node.source` would be equal to
`\" o`

unittests/accents.py
```diff
@@ -0,0 +1,8 @@
+from plasTeX.TeX import TeX
+
+def test_accent():
+    input_data = r"""\"o"""
```
PatrickMassot, line 4:
```suggestion
    input_data = r'\"o'
```
That should be enough "

PatrickMassot:
Ok, I see now. Thanks. I suggested a very minor stylistic improvement. Otherwise this is good to go.

dalcde:
It should be good now

PR #174, opened 2020-04-17
Deal with latex failures

### LINKED ISSUE #139: dealing with `latex` failures: redirect input  and `nonstopmode`
hi

AFAIU in `plasTeX/Imagers/__init__.py` , the `latex` program is used to transform LaTeX code to images. But if the latex compilation fails, then `latex` may stop asking for further instructions; at the same time, its output is captured parsed, so the user has no knowledge that this is happening, and `plastex` stops forever.

A simple way to solve this problem is to redirect the input as follows

```
--- a/plasTeX/Imagers/__init__.py
+++ b/plasTeX/Imagers/__init__.py
@@ -614,6 +614,7 @@ class Imager(object):
         p = subprocess.Popen(shlex.split(cmd),
                      stdout=subprocess.PIPE,
                      stderr=subprocess.STDOUT,
+                     stdin=open(os.devnull),
                      universal_newlines=True)
         cmd_line = '{} {}{}images.tex'.format(program, tempdir, os.sep)
         while True:
```

A simple way to test the effect of this patch is to use `plastex` on this LaTeX file

```
\documentclass{article}
\usepackage{foobarmatic}
\begin{document}
\[ \cos(x)=1\]
\end{document}
```

Note that there are other places where `subprocess` is used : I would advise that, everywhere where user input is not not expected, `stdin` should be redirected .

Another advice: when `latex` is called but it is not interactive, then the option `-interaction=nonstopmode` should be used.

bye . a.

dalcde:
Fixes #139. This causes plastex to not hang/wait for user input if
`latex` fails when called by the imager.

Adding a test for this turns out to be very tricky, since pytest seems
to automatically redirect stdin to /dev/null , so I cannot reproduce the
error. Someone more knowledgable about pytest might be able to fix this,
but otherwise I'll leave this as is.

### DIFF
--- a/plasTeX/Imagers/__init__.py
+++ b/plasTeX/Imagers/__init__.py
@@ -621,6 +621,7 @@ class Imager(object):
         p = subprocess.Popen(shlex.split(cmd),
                      stdout=subprocess.PIPE,
                      stderr=subprocess.STDOUT,
+                     stdin=subprocess.DEVNULL,
                      universal_newlines=True)
         cmd_line = '{} {}{}images.tex'.format(program, tempdir, os.sep)
         while True:
--- a/unittests/LaTeXError.py
+++ b/unittests/LaTeXError.py
@@ -1,0 +1,34 @@
+import os, subprocess
+from pathlib import Path
+
+def test_latex_imager_failure(tmpdir, pytestconfig):
+    """While rendering using LaTeX in an imager, a missing package shouldn't
+    block plasTeX."""
+    tmpdir = Path(str(tmpdir))
+
+    (tmpdir / "CustomRenderer").mkdir()
+    (tmpdir / "CustomRenderer" / "__init__.py").write_text(r"""
+from plasTeX.Renderers import Renderer as BaseRenderer
+
+class Renderer(BaseRenderer):
+    def  __init__(self, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


PatrickMassot:
Yes, that one is tricky to test. It seems you can try:
```python
import os, subprocess                        
from pathlib import Path                                     
                                                             
def test_latex_imager_failure(tmpdir, pytestconfig):         
    """While rendering using LaTeX in an imager, a missing package shouldn't  
    block plasTeX."""                                        
    tmpdir = Path(str(tmpdir))  
    (tmpdir/'test.tex').write_text(r"""  
    \documentclass{article}  
    \usepackage{foobarmatic}  
    \begin{document}  
    \[ \cos(x)=1\]  
    \end{document}""")  
  
    cwd = os.getcwd()  
    os.chdir(str(tmpdir))                                                                                 
    capmanager = pytestconfig.pluginmanager.getplugin('capturemanager')  
    capmanager.suspend_global_capture(in_=True)            
    proc = subprocess.Popen(['plastex', '--renderer=XHTML', 'test.tex'])      
    proc.communicate(timeout=2)                                               
    capmanager.resume_global_capture()      
```

dalcde:
I'm closing this in favor of a more thorough cleanup of the imager.

Issue #170, opened 2020-04-16
Mathjax CDN changes


PatrickMassot:
The default values in the HTML5 renderer need to take https://www.mathjax.org/cdn-shutting-down/ into account.

PatrickMassot:
Fixed in  7bbce77

PR #175, opened 2020-04-19
Update default HTML5 MathJax url

dalcde:
The existing default MathJax url redirects us to the new default url,
except it hardcodes version 2.7.1 whereas the new default always points
to latest, cf #170


PatrickMassot:
Grmph, I missed the fact that you didn't update the functional tests. I really need to understand why Travis is not reporting in the PR pages the work it reports at https://travis-ci.org/github/plastex/plastex/pull_requests. I love tests but I hate Travis...

PatrickMassot:
Fixed in ed9671d

PR #176, opened 2020-04-19
Remove bs4 dependency

dalcde:
This was previously used for the tikz renderer which has now been
replaced. It is still in requirements_tests.txt since it is needed for
`unittests/Packages/conftest.py`


PatrickMassot:
Good catch!

PR #178, opened 2020-04-19
More robust imager implementation

### LINKED ISSUE #139: dealing with `latex` failures: redirect input  and `nonstopmode`

dalcde:
This started off as an attempt to fix the following issue, but the
fix ended up requiring rewiring a lot of the internals of Imager, and
I tidied up a lot of the code along the way.

The Imager puts every image in a plasTeXimage environment, which in
particular inserts a pagebreak. It then assumes the nth image is on the
nth page. This assumption, however, is not always valid. For example,
someone might have put an `\AtBeginDocument` in the preamble, which
would shift all the contents. Another possibility is that some of the
contents are so long that they span more than one page (this will cause
other issues that are not addressed by this fix).

The solution is for the plasTeXimage environment to write the current
page number and destination filename into a csv file, which is then read
by the imager. The imager then uses this information to copy the image
files to the right destination. In addition to fixing the aforementioned
issue, this is also much less fragile. Previously, we had to carefully
keep the converted images and destination filenames sorted so that they
correspond to each other. In the new code, we maintain an explicit
bijection between them all along.

This comes along with some under-the-hood changes. Previously, we
maintained two separate temporary directories; one for compiling latex
and the other for converting images. Since the two processes now share
data, it makes more sense to use the same temporary directory for both
of them. This also makes debugging easier.

As part of the rewriting, the imager is now more eager to bail when it
encounters errors. In my experience, such errors usually lead to
erroneous results, and it is better to throw an error than hoping that
things still work out fine. Currently it doesn't halt the entire
plasTeX compilation process; it just doesn't attempt to scrape images
out of whatever is remaining. The temporary directory is never cleared
when an error is encountered, and the user will be told of the location.

To ensure I didn't break anything, I expanded the imager unittest to
test all imagers but OSXCoreGraphics.py , and tidied up some of the
code. The test is also modified to inject something at the first page
with `\AtBeginDocument` to ensure the issue is in fact fixed. At first,
I wanted to make this a separate test, but at the end it turns out to be
exactly the same test except with an `\AtBeginDocument`, so I merged the
two.

Minor change: Originally, the latex is compiled with \scrollmode. This
restricts possiblity of user interaction, but does not stop it
completely. Here we replaced it with \nonstopmode, which fixes #139. We
also redirect stdin to /dev/null so we now fix it twice.

I acknowledge this is a big commit with many changes, but they are hard
to separate from each other, and even if I managed, each of the
individual changes requires updating all the existing imagers, which
seems pointless.


PatrickMassot:
Did you test whether this has any impact on performance? Converting images is done a *lot* by renderers that rely on it for math rendering.

dalcde:
I have not done any profiling, but I don't expect there to be much
significant impact.

My experience is that most of the time is spent running the
pdf/dvi-to-image converter. There are two kinds of such converters:

  - The first kind coverts all pages to images in a single command. In
    the new code, it does the same, and then picks out the pages that are
    relevant.

  - The second kind converts the pages one by one. The new code also
    converts them one by one, but only converting the pages that are
    relevant (of course, if the old code was working at all, this would
    be all pages).

In either case, the slowest part of the code is unaffected. I don't
think it costs much for the latex code to write the data to the csv, and
the python glue should be very fast compared to everything else.

unittests/imagers/imagers.py
```diff
@@ -16,28 +16,42 @@
...
 
     tex = TeX()
     tex.ownerDocument.config['images'][kind] = imager
-    tex.ownerDocument.config['images']["compiler"] = compiler
+    if compiler is not None:
+        tex.ownerDocument.config['images']["compiler"] = compiler
+
     tex.input(r'''
     \documentclass{article}
+    \AtBeginDocument{You should not be getting this. If you are, you are doing something wrong.}
```
PatrickMassot, line 44: Could you expand on that mysterious message?

PatrickMassot:
Could you please rebase and resolve conflicts?

You added tests showing that imagers still work as they used to be, but it seems to me they don't prove you improved anything. Could you try to add test that currently fail on master and work with this patch?

dalcde:
The new test fails on master (note that the only substantial change is the `\AtBeginDocument` line, which was why I was reluctant to separate it out; maybe I should)

dalcde:
I rebased and reworded the `\AtBeginDocument` comment.

dalcde:
The tests seem to be failing catastrophically...

dalcde:
Everything working now

PatrickMassot:
> Everything working now

This is not Travis' opinion: https://travis-ci.org/github/plastex/plastex/pull_requests

dalcde:
It's now actually working.

PatrickMassot:
Great!

Issue #90, opened 2017-08-08
renewcommand fails for builtin command


PatrickMassot:
Currently ``\renewcommand`` works only for user defined commands but not builtin commands. Consider the following input:
```latex
\documentclass{article}
\renewcommand{\textbf}[1]{YOP}
\begin{document}
\textbf{Test}.
\end{document}
```
When complied by pdflatex, you see YOP. But plasTeX (master branch, default config) displays **Test**.
This is due to [these lines](https://github.com/tiarno/plastex/blob/f07aab870059c1b548e54c0a7b17482884cc9060/plasTeX/Context.py#L945-L946).

kesmit13:
Yes.  This is by design.  I've always made the assumption that if there is a Python-implemented class to handle a macro, that it was hand-tuned to work better than a LaTeX implementation of that same macro.  So I don't allow a LaTeX macro to redefine a Python class-based macro.

PatrickMassot:
I see. I won't discuss that design choice, but I think it should be documented somewhere (it took me a *long* time to understand what was going on). You could also mention that using ``\def`` is not forbidden in this context.

kesmit13:
I'll admit plasTeX gets into some weird areas when trying to merge the worlds of (La)TeX macros vs. Python class, linear documents vs. DOM objects, etc.  Things don't always fit together perfectly.  I don't recall why ``\def`` was handled differently though.

dalcde:
I also find this behavior counter-intuitive. When I use `\renewcommand`, I almost always mean "I want to define a command with this name. There is already a command that uses this name but I never use it, so let me overwrite it", as opposed to "I have a better definition of this command that performs the same function".

I propose the following behaviour:
1. `\newcommand` *never* overwrites any existing definition. If there is an existing python definition, we print a warning and keep the python definition. If there is an existing latex definition, it throws an error. This avoids the problem in #22 .
2. `\renewcommand` by default always overwrites any existing definition.
3. A python implementation that doesn't want to be overwritten can set a `forceKeep` parameter to avoid being overwritten by `\renewcommand`.

The main problem I can foresee is that some packages may want to redefine a command to be compatible with their package, e.g. `amsthm` redefines `newtheorem`. However, in my experience, the packages are usually sufficiently complex that we will usually have a python redefinition of the package and the latex will not be parsed.

PatrickMassot:
@kesmit13 Do you mind if we follow @dalcde 's suggestion?

[...]

PR #180, opened 2020-04-19
Make \let less broken

dalcde:
This is still somewhat broken but less broken.

I believe this works correctly if we \let\foo\bar where \foo and \bar
are both escape sequences.

The correct behaviour for \let\foo=x seems to be extremely complicated
when x is not an escape sequence. For example, if we set \let\foo={,
then apparently \foo can sometimes take the role of { but not always.
Currently, we simply replace \foo with { in the Tokenizer in this case
(which was the previous behaviour for *all* \let definitions). This is
incorrect but just as wrong as the previous version.

One particular problem with this is that we cannot use let twice, i.e.

\let\foo=a
\let\foo=b

fails to define \foo as b, as the second line gets turned into \let a=b.
This is not something special about let. This is the same "\foo is
sometimes `a` but not always" problem.

Without the commit, the first, third and fourth tests fail. The second
test passes but only by coincidence.

### DIFF
--- a/plasTeX/Context.py
+++ b/plasTeX/Context.py
@@ -2,6 +2,7 @@
 
 from plasTeX import ismacro, macroName
 from plasTeX.Logging import getLogger
+from plasTeX.Base.TeX.Primitives import relax
 from plasTeX.Tokenizer import Tokenizer, Token, DEFAULT_CATEGORIES, VERBATIM_CATEGORIES
 import os
 import configparser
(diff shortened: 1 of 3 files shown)


dalcde:
This hits #90 again. If we don't make `\relax` a special case, then if we try to do
```latex
\let\foo\relax
\newcommand\foo{...}
```
then the second `\newcommand` will not register, since we are trying to override a macro with python implementation.

PatrickMassot:
Thanks!

PR #181, opened 2020-04-19
Specify python3 explicitly

dalcde:
We no longer support python2

### DIFF
--- a/plasTeX/Base/LaTeX/Accents.py
+++ b/plasTeX/Base/LaTeX/Accents.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 """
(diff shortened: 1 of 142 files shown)


PatrickMassot:
Do you understand the purpose of all those lines? Almost of these files are not meant to be executed but only imported. Why not removing those lines entirely?

dalcde:
I agree. I merely did a sed -i **/*.py; easier than handpicking lines to
remove. Would you like me to remove the lines as well?

PatrickMassot:
If you have the patience (or want to play with scripting) then yes.

PatrickMassot:
Thanks!

PR #182, opened 2020-04-20
Fix counter presentation

### LINKED ISSUE #170: Mathjax CDN changes

dalcde:
Previously, TheCounter yanks any counter value that is 0. This is the
desired behaviour in classes such as \thefigure in book or report, which
wants to convert Figure 0.1 to Figure 1. However, this is generally
undesirable. It turns subsection 0.1 to subsection 1, and even worse,
4.0.3 to 4.3.

This commit modifies the behaviour in two ways. Firstly, the yanking
behaviour is opt-in. Secondly, it only removes instances of "0." at the
beginning of the counter value, instead of everywhere.

Fixes #170

### DIFF
--- a/plasTeX/Context.py
+++ b/plasTeX/Context.py
@@ -779,7 +779,7 @@ class Context(object):
         """
         self.contexts[-1].categories = self.categories = VERBATIM_CATEGORIES[:]
 
-    def newcounter(self, name, resetby=None, initial=0, format=None):
+    def newcounter(self, name, resetby=None, initial=0, format=None, trimLeft = False):
         """
         Create a new counter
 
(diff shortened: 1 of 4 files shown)


PatrickMassot:
This looks great, but we also need tests that check counters are *not* trimmed when they shouldn't be.

dalcde:
On Sat, Apr 25, 2020 at 06:35:02AM -0700, Patrick Massot wrote:
>This looks great, but we also need tests that check counters are *not* trimmed when they shouldn't be.

That's checked in the last line of the test --- trimLeft is not set for
section and it checks that we get 0.1 correctly.

PatrickMassot:
I meant checking  the almost same source but with another document class that doesn't trim counters.

dalcde:
The other document classes (`article`) do not number figures by section, so we just get 1, 2, 3, etc.

PR #183, opened 2020-04-20
Removed Token classes that can never be initialized

dalcde:
They previously existed because Tokenizer.iterchars wanted to return
Token objects, but we can just return the catcode and character. From a
TeX point of view, an Active character is a valid token, but we decided
to encode them as special EscapeSequence's instead, so we can eliminate
the class as well.

The Tokenizer.iterchars function is only used in two places. The first
is the Tokenizer iteration function, which is modified accordingly. It
is also used in TeX.iterchars, which was introduced in cefff9c, but its
use was eliminated in b1d76f9. The function doesn't really make much
sense in the first place and I believe that is why we stopped using it.
I deleted this function completely.

This commit came from thinking about what we should do when someone tries to `\let` a command to be a comment character, e.g. `\let\foo=%`. Of course, this doesn't work, since `%` is thrown away by the Tokenizer so it never reaches the `\let` command. Removing these classes that can never exist emphasizes the fact that in a correct implementation of `\let`, we never have to worry receiving `Comment`.

### DIFF
--- a/plasTeX/DOM/__init__.py
+++ b/plasTeX/DOM/__init__.py
@@ -1746,16 +1746,6 @@ class Text(CharacterData):
         return self.parentNode.textContent
 
 
-class Comment(CharacterData):
-    """
-    Comment
-
-    http://www.w3.org/TR/2004/REC-DOM-Level-3-Core-20040407/core.html#ID-1728279322
-    """
-    nodeName = '#comment'
-    nodeType = Node.COMMENT_NODE
-    __slots__ = Node.TEXT_SLOTS
-
 
 class TypeInfo(object):
     """
@@ -2024,7 +2014,6 @@ class Document(Node):
     elementClass = Element
     documentFragmentClass = DocumentFragment
     textNodeClass = Text
-    commentClass = Comment
     cdataSectionClass = CDATASection
     processingInstructionClass = ProcessingInstruction
     attributeClass = Attr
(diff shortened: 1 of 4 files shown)


dalcde:
On Sat, Apr 25, 2020 at 07:06:37AM -0700, Patrick Massot wrote:
>
>I'm a bit uneasy with that comment removal thing. Are you sure you are not throwing away information?

I just removed the comment class. It is never instantiated. This is
consistent with what what TeX does (comments never get past the
tokenizer).

>While you're adding type annotations, don't hesitate to go as far as using [NewType](https://mypy.readthedocs.io/en/stable/more_types.html#newtypes). This is a textbook case: the `int` annotation is extremely vague while you have a very specific `CategoryCode` in mind (we could also use `Enum` here but this is probably not worth the effort). Of course the `NewType` declarations have to go in the root files.

Good point here.

>It looks like you announced a tuple and return a list. Is mypy happy about this?
For some reason I thought tuples in python are arrays... I'll fix this.

>> @@ -431,34 +406,34 @@ def __iter__(self):
>                 # Get name of command sequence
>                 self.state = STATE_M
>
>-                for token in charIter:
>+                for [c, t] in charIter:
>
>Could you please avoid single letter variable names?

I could try to come up with something, but I'm really short on variable
names now (with c_ and t_ in the further inner loop)

PR #185, opened 2020-04-20
Add option disable some charsubs

dalcde:
It provides only marginal typographical benifits, and breaks search for
words with apostrophes.

Fixes gerby-project/plastex#37

### DIFF
--- a/Doc/command.tex
+++ b/Doc/command.tex
@@ -148,6 +148,16 @@ table of contents.  By default, only sections that create files will show
 up in the table of contents.
 \end{configuration}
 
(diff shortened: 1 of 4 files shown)


PatrickMassot:
I don't really like this. I think the typographical benefit is very real. LaTeX certainly doesn't use the dactylographic apostrophe. But I would merge a PR introducing an apostrophe opt-out option.

dalcde:
I'm now happy with the current implementation.

PatrickMassot:
Yes, this is much better. Thanks!

PR #187, opened 2020-04-20
Add \romannumeral command

dalcde:


plasTeX/__init__.py
```diff
@@ -1510,6 +1510,46 @@ def addtolength(self, len):
     def new(cls, *args, **kwargs):
         return muglue(*args, **kwargs)
 
+def numToRoman(x: int) -> int:
```
PatrickMassot, line 1513: Again this type annotation is wrong, you are returning a `str`. Are you actually running any check here?

dalcde: Unfortunately not. I can't get mypy to run without throwing errors on every unannotated thing... I should figure this out one day.

PatrickMassot: Weird. mypy is very explicitly designed with this situation in mind, allowing gradual adoption. This is worth investigating or else you should stop adding type annotations because wrong annotations are worse than no annotation.

dalcde: There are errors of various kinds, many of which are because the defaults are set to None so mypy thinks it is a None type. There is supposed to be a flag that suppresses it but it doesn't seem to work. If you have a working configuration please let me know.

In any case, I think annotating types without checking is better than describing them in words in the documentation, which can just as equally be wrong, and I think most people do not object to the latter. Of course, wrong documentation is harmful, but this just means we should be careful in checking by hand that we annotated the right type (which apparently I failed to).

dalcde: I can also run mypy and check that the errors are not because of the new type annotations, which I've done for some of my new changes.

PR #190, opened 2020-04-21
Fix expandafter implementation

dalcde:


unittests/helpers/utils.py
```diff
@@ -30,3 +33,32 @@ def __init__(self, message):
 
     # out.stderr is a byte sequence, but float can take byte sequences
     return float(diff)
+
+# This checks whether plastex produces the same output as tex. This only works
+# for things with simple plain text output, and is intended to be used for
+# testing primitives such as \let, \def, \expandafter etc.
```
PatrickMassot, line 39: Why not using this as a function docstring?

plasTeX/__init__.py
```diff
@@ -143,6 +143,15 @@ class Macro(Element):
     # Force there to be at least on paragraph in the content
     forcePars = False
 
+    # Macros do not have but we give EscapeSequence a catcode of 0 so I guess
```
PatrickMassot, line 146: Is there a missing word here?

dalcde:
On Sat, Apr 25, 2020 at 09:30:14AM -0700, Patrick Massot wrote:
>
>Unfortunately this is beyond my TeX macro skills. I hope the existing tests are good enough to ensure you didn't break anything.

The previous implementation was completely broken because on line 473 it
said `nextok` instead of `nexttok`. It has never been working; if it
did, it was by pure coincidence.

PatrickMassot:
I see. That typo does look very embarrassing, and certainly suggests this code was never used anywhere. So I don't see any reason not to merge.

PR #194, opened 2020-04-24
Copy xymatrix implementation from gerby.

dalcde:
I have not verified the implementation but added tests to check that
source returns the right result.

### DIFF
--- a/plasTeX/Packages/xy.py
+++ b/plasTeX/Packages/xy.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python
+
+from plasTeX import Command
+
+class xymatrix(Command):
+    args = '{ str }'
+    doCharSubs = False
+
+    class EndRow(Command):
+        """ End of a row """
+        macroName = '\\'
+
+    class CellDelimiter(Command):
+        """ Cell delimiter """
+        macroName = 'active::&'
+        pass
+
+    class ar(Command):
+        args = '[ misc:str ]'
+
+    class arAt(Command):
+        macroName = "ar@"
+        args = 'arType:str [ misc:str ]'
+
+        def postArgument(self, arg, value, tex):
+            self.argSource = self.argSource.replace("~ ","~")
+            super(Command, self).postArgument(arg, value, tex)
+
+    class omit(Command):
+        @property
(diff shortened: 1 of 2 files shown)


PatrickMassot:
In the PR title, do you mean Gerby instead of plasTeX?

dalcde:
Yes...

PatrickMassot:
Do you know if Gerby would work if you used the verbatim environment you used for TikZ? Otherwise I'm fine merging without really understanding what Gerby does with this. We don't want to spend too much time on this super-deprecated package.

pbelmans:
I agree `xy` is super-deprecated, but given how important it is for the Stacks project and Kerodon, it is unfortunately important.

I don't know how the implementation you refer to works, but XyJax needs the raw TeX code, maybe with macros expanded, whatever I did in Gerby is trying to do exactly this, if I remember correctly (but it's been 3 years or so that I've done this).

PatrickMassot:
@pbelmans it would be really really helpful if you give us things to test here.

dalcde:
On Sat, Apr 25, 2020 at 08:53:26AM -0700, Patrick Massot wrote:
>Do you know if Gerby would work if you used the verbatim environment you used for TikZ? Otherwise I'm fine merging without really understanding what Gerby does with this. We don't want to spend too much time on this super-deprecated package.

This doesn't quite work because you usually invoke \xymatrix inside
math mode, so it is hard to automatically deploy the imager in this
case. If we just put everything in verbatim without using the imager,
there would be no macro expansions at all, which would break a lot of
things (e.g. if we defined \R to \mathbb{R}, we don't want \R to show
up).

As Pieter said, the goal of this package is to make sure running
`obj.source` returns a source that mathjax understands. This generally
involves macro expanding all user-defined macros but not "built in" xy
matrix macros. This largely involves listing out all xy matrix macros to
prevent plasTeX from doing anything with it (and sometimes control the
spacing precisely), but there is not much programmatic things to do.

PatrickMassot:
Ok, this is what I imagined. Since this seems good enough for Gerby let's move on (and hope people will stop using that zombie thing really soon).

PR #197, opened 2020-04-26
Fix or suppress all mypy errors; add mypy to test

dalcde:
The number of errors is now down to 64.

Most of these are programming errors that are detected by mypy. The fix is intended to reproduce existing behaviour while fixing errors, without regards to whether existing behaviour is sane or correct.

Many errors are caused by
 - https://github.com/python/mypy/issues/5080
 - https://github.com/python/mypy/issues/4125

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,3 @@
+[mypy]
+ignore_missing_imports = True
+allow_untyped_globals = True
--- a/plasTeX/Base/LaTeX/Accents.py
+++ b/plasTeX/Base/LaTeX/Accents.py
@@ -7,7 +7,7 @@ C.3.4 Accents and Special Symbols (p173)
 
 from plasTeX import Command
 from plasTeX.DOM import Node, Text
-
+from typing import Optional
 
 #
 # Table 3.1: Accents
@@ -47,7 +47,7 @@ class Accent(Command):
 
 
 class Grave(Accent):
-    macroName = '`'
+    macroName = '`' # type: Optional[str]
     chars = {
         'A': chr(192),
         'E': chr(200),
@@ -65,7 +65,7 @@ class Grave(Accent):
     combining = '\u0300'
 
 class Acute(Accent):
-    macroName = "'"
+    macroName = "'" # type: Optional[str]
     chars = {
         'A': chr(193),
         'E': chr(201),
@@ -97,7 +97,7 @@ class Acute(Accent):
     combining = '\u0300'
 
 class Circumflex(Accent):
-    macroName = '^'
+    macroName = '^' # type: Optional[str]
     chars = {
         'A': chr(194),
         'E': chr(202),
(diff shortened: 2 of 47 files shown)


PatrickMassot:
This PR makes me very nervous, although I love the intend. The repeated things issue is really weird. Do you understand what could have been the purpose of all this? Are you sure the existing tests cover everything you removed? This PR could create problems that would be very difficult to debug in a couple of months.

PatrickMassot:
Also, why did you remove stuff like `class uparrow(MathSymbol): str = chr(8593)`. Is this incorrect, or not rendered correctly in web browsers?

dalcde:
The stuff like uparrow is repeated. The first definition is what you
listed, while the second definition is a no-op that lets mathjax/katex
render it. Since the second overwrites the first, I removed the first.
(I also believe it leads to better results, but that's of secondary
concern for this PR)

On Sun, Apr 26, 2020 at 06:30:16AM -0700, Patrick Massot wrote:
>This PR makes me very nervous, although I love the intend. The repeated things issue is really weird. Do you understand what could have been the purpose of all this? Are you sure the existing tests cover everything you removed? This PR could create problems that would be very difficult to debug in a couple of months.
I think they are mostly oversights, since these commands were added
systematically based off some reference and they didn't check for
overlap. For example, the TeX version of the raggedright command was
introduced in commit d575a87ee in TeX while the LaTeX version was
introduced in c8b7d2bbaa, both of which add huge lists of commands.

Potentially, one might be able to use plasTeX to deal with plain TeX
documents in which case we only want Base/TeX and not Base/LaTeX, but I
don't think that is something we do/want to support. It also wouldn't be
correct because strictly speaking, LaTeX is not a superset of TeX; they
can implement commands differently.

PatrickMassot:
There is an advantage to converting many commands to unicode rather than letting mathjax rendering them from the macro name. When mathjax is executed on client side, you first see the unrendered  code while mathjax is working. This lead to a lot of HTML reflow which is visually very distracting. Whereas unicode takes approximately the same space before and after mathjax rendering. In some pages which mostly use isolated symbols this can make a huge difference.

dalcde:
Sure. MathJax also automatically turns these unicode into the mathjax
rendered thing. However, as I said, the goal of this commit is merely to
replicate old behaviour. There are too many individual items for me to
decide which is better one by one.

dalcde:
A more relevant point is that setting `str = chr(...)` doesn't actually cause the HTML5 renderer to produce the unicode character in the output, because math is renderered using the `source` property.

PatrickMassot:
I'm not sure that running `mypy` with all python versions is really helpful. We could have a tox mypy environment like:
```
envlist = py35, py36, py37, py38, mypy
...
[testenv:mypy]
basepython = python3.8
deps = mypy
setenv = MYPYPATH={toxinidir}
commands = mypy plasTeX unittests
```
but running mypy is very fast compared to our other tests anyway. So I only pasted the above code for future reference if needed.

PR #198, opened 2020-04-26
Use compare_output in more unittests

dalcde:
This catches a few previous tests that are actually incorrect.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ htmlcov/
 .cache
 nosetests.xml
 coverage.xml
+unittests/helpers/compare_cache/
 
 # Translations
 *.mo
--- a/plasTeX/Base/TeX/Primitives.py
+++ b/plasTeX/Base/TeX/Primitives.py
@@ -203,7 +203,7 @@ class ifnum(IfCommand):
     def invoke(self, tex):
         self.parse(tex)
         attrs = self.attributes
-        attrs['b'] = tex.readNumber(optspace=False)
+        attrs['b'] = tex.readNumber()
         relation = attrs['rel']
         a, b = attrs['a'], attrs['b']
         if relation == '<':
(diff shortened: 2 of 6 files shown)


PatrickMassot:
I'm a bit worried about generalizing the use of `compare_output`. This will make the test suite very slow for no good reason since we don't expect the pdflatex output to change. I like the idea of comparing output with pdflatex, but the pdflatex output should be cached, using something like the benchmark framework setup for functional and imager tests. You could even keep the same code in tests, but simply modify `compare_output` so that it populates a `unittests/compare_cache` folder, and run `pdflatex` only if the corresponding cache does not exit. That cache folder could be left out of the repository so Travis would catch pdflatex variations, but developers wouldn't need to wait for pdflatex.

Having many slow tests is dangerous because it discourages you and me to run `pytest` or `tox` before pushing. We don't care if Travis is a bit slow (although this can be painful if it takes one hour, especially when debugging what happens there).

dalcde:
Given that the change revealed a few incorrect tests, I think we should
use compare_output more often. On the other hand, I do notice a slowdown
in running thests.

I think it makes sense to have a local cache (that is not committed),
but I am not against including the cached results as well.

PatrickMassot:
Yes, this is much better. Now, what's the matter with all those xfail tests? Do you understand how to fix them?

dalcde:
I tried to include some brief comments for each of them.

One of them tests \ifeven, which we implemented but isn't an actual TeX
command. The others are basically unimplemented. They are just
hardcoded to be True or False.

PatrickMassot:
I don't understand the point of replacing the `\ifeven` test. If this is purely plasTeX then why do you want to compare with `pdflatex`? Can't we keep the old test?

For the incorrect ones, I would much prefer if you could try to fix them before we merge. But if you don't have time for this, or are not interested, then pytest will remind us this issues exist.

dalcde:
On Mon, May 04, 2020 at 03:05:59AM -0700, Patrick Massot wrote:
>I don't understand the point of replacing the `\ifeven` test. If this is purely plasTeX then why do you want to compare with `pdflatex`? Can't we keep the old test?
I think the correct thing to do is to remove the implementation.

>For the incorrect ones, I would much prefer if you could try to fix them before we merge. But if you don't have time for this, or are not interested, then pytest will remind us this issues exist.

I think these are pretty difficult to implement.

PatrickMassot:
This was mostly waiting for a decision on #215 since the actual failure seem hard to fix and non-urgent.

PR #204, opened 2020-05-05
Add partial cleveref implementation

dalcde:


PatrickMassot:
Thanks, this is really good to have!

PR #207, opened 2020-05-06
Fix mypy error from 8c9df7e

### LINKED ISSUE #201: Config Manager
@dalcde I just reverted your commit 99fc05c since it completely broke config file reading. 

There is a recurring theme here. The `ConfigManager` subpackage is cursed:
* This is 3000 lines of python which have *nothing* to do with either TeX or DOM.
* It has been a major pain when switching to python3.
* I think it caused most of the `mypy` PR pain.
* it relies a lot on side effects (as illustrated by your breaking change)
* It has things like [GenericParser](https://github.com/plastex/plastex/blob/525a41ed299456db1c7f095ce3f75555dc196d3d/plasTeX/ConfigManager/Generic.py#L30) which is a class with no parent and no `__init__`  whose methods [refer](https://github.com/plastex/plastex/blob/525a41ed299456db1c7f095ce3f75555dc196d3d/plasTeX/ConfigManager/Generic.py#L60) to things like `self.optional`, hoping that subclasses will define it.

My hope is no alternative existed 20 years ago when this was written, but maybe we could get rid of it now. I say: let's investigate.

dalcde:


PatrickMassot:
It seems the PR text is unrelated to the PR content, probably through copy-paste mistake. Could you fix this?

dalcde:
This is correct. mypy complains that label.counter is an Optional[str]
instead of str, but the return type of the function requires a str. By
testing if it is None, we avoid this issue.

PatrickMassot:
I still don't see the relation between this and your description:
```quote
Anything not in ConfigManager/init.py is not used; it is kept for
reference before I complete. There were some changes elsewhere that were exploratory and irrelevant

Fixes #201
```
which literally a subset of the description of #206.

dalcde:
Ah sorry I see what you mean. Yeah that is a copy-and-paste error.

PatrickMassot:
I think the new code is confusing. If you claim that `if label.counter is not None:` is actually useless and you wrote it only to satisfy mypy then you should either remove the [else branch](https://github.com/plastex/plastex/blob/bb9b6cb8fa6759e8443182b510c55552cb58ade8/plasTeX/Packages/cleveref.py#L26-L28) since [this line](https://github.com/plastex/plastex/blob/bb9b6cb8fa6759e8443182b510c55552cb58ade8/plasTeX/Packages/cleveref.py#L34) will do it anyway or, probably clearer, you should put in the else branch
```python
assert label.counter
return label.counter
```
which clearly asserts what you think should happen and will make debugging much easier if something goes wrong.

dalcde:
I agree this is a better solution. I've updated the commit.

PatrickMassot:
Thanks!

PR #215, opened 2020-05-10
Remove non-existent \ifeven command

dalcde:


PatrickMassot:
What would we gain from removing this?

dalcde:
We get rid of code that is arguably wrong

PatrickMassot:
What I don't understand is why this is "wrong". The test suggests it works.

dalcde:
A TeX engine should not implement commands that do not exist in tex.

One possible (albeit unlikely) issue is that this prevents people from defining their own command called \ifeven.

PatrickMassot:
I can't say I'm really convinced but I merged it since it seems important to you, and I hope nobody else cares.

Issue #200, opened 2020-05-04
Add option to disable loading TeX packages


dalcde:
This would be very helpful since these fail most of the time anyway. There ought to be a way to manually force loading a particular package (possibly in the TeX source) in case some can actually be parsed (most likely written by the user themselves).

In such situations, if there is no python implementation, the package is simply not loaded, and a warning is produced. This feature would be especially useful when users use some large preamble that is included by all their documents, and many of the packages might not actually be used in the particular document they try to compile.

PatrickMassot:
I agree this would be useful. I guess the list of packages to try to load anyway can be a plasTeX option.

PatrickMassot:
Fixed in e23616a11f20989072997f3049f8a8a87f266224

PR #217, opened 2020-05-10
Extremely primitive imakeidx implementation

dalcde:


PatrickMassot:
Do you actually want this implementation, or is it only waiting for #200?

dalcde:
It lets makeindex take an optional argument, while the original one doesn't. It doesn't quite matter in practice because it is in the preamble, so the leftover argument wouldn't do anything, but it may lead to unexpected issues in the future

PR #221, opened 2020-05-20
Don't always generate bitmap images

dalcde:
Currently, if one requests a VectorImage from the imager, a bitmap image
is also produced to read off the desired height and width. These
properties are not used in certain renderers such as HTML5. Moreover,
producing these bitmaps is extremely slow.

This commit allows renderers to decide whether they want these bitmaps
to be produced, and disables it for HTML5.


PatrickMassot:
Did you check that the documentation doesn't need any update here?

dalcde:
I've updated the documentation.

PatrickMassot:
Thanks!

PR #224, opened 2020-07-21
Don't include empty properties in CSSStyles.inline

### LINKED ISSUE #223: CSSStyles.inline can produce empty properties
`obj.style.inline` can produce properties with empty values if the node for the value produces empty text.

For example, `\includegraphics[width=\textwidth]{image.png}` produces `style="width:"` in the `` tag in the HTML5 renderer.

christianp:
fixes #223

### DIFF
--- a/plasTeX/__init__.py
+++ b/plasTeX/__init__.py
@@ -102,9 +102,10 @@ class CSSStyles(dict):
         string containing inline CSS
 
         """
-        if not self:
+        nonempty_keys = {k:str(v) for k,v in self.items() if v is not None and str(v)!=''}
+        if not nonempty_keys:
             return None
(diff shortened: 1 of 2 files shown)


PatrickMassot:
Could you please add a couple of unit tests (one showing the fix and one proving it still works when something should be written)?

christianp:
I've added some tests. I hope I've followed your style of testing!

PatrickMassot:
I'm sorry I missed this message in August. Thanks for adding the tests. The question about testing style unfortunately doesn't make any sense because plasTeX has several styles of testing. I'll reformat this to the current style.

PatrickMassot:
FYI, the new style can be seen in 8648fa13150b0794b41b6573325cccf418a4ece3. Hopefully this diff makes it clear why we switched to pytest for new tests.

Issue #179, opened 2020-04-19
\ifplastex incompatible with imager


dalcde:
Consider the following code (with XHTML renderer; just looking for something convenient to trigger the imager)
```latex
\documentclass{article}
\newif\ifplastex
\plastexfalse

\ifplastex
  \newcommand\foo{a}
\else
  \newcommand\foo{b}
\fi
\begin{document}
$\foo$
\end{document}
```
One would expect to get `b` since this is compiled by the `LaTeX` compiler. However, we get `a`. This is since the imager doesn't actually use the actual preamble of the document, but reproduces the preamble using the parsed DOM. However, the `source` function is not always faithful, and indeed when we invoke `\if`, `plastex` simply discards the other branch instead of remembering the `\if`.

PatrickMassot:
I think the semantic of `\ifplastex` is unclear here, one could argue we are indeed using plasTeX hence should get `a`. I'm not sure this should be "fixed", but we could warn users in the documentation.

To be honest I never used this `\ifplastex` trick. I always use plasTeX with (at least) three files:
`content.tex`, `web.tex`, and `paper.tex` where `web.tex`, and `paper.tex` both contain `\input{content}` but have different preambles and are compiled by `plastex` and `xelatex` respectively.

dalcde:
I've used \ifplastex if there is some command that plasTeX couldn't
parse correctly. One concrete example is that I want to run the
following code:
```
\let\index\relax
\DeclareMathOperator\index{index}
```
before the \let PR, plasTeX would not understand the first line and
throws an error. I then wrapped the first line in \ifplastex, which
happened to be fine since plastex didn't mind me redefining index, but
latex did.

In the imager code, I definitely want the `\let\index\relax`, or else
the code would not compile.

dalcde:
A related issue occurs when \input is used in the preamble. The produced source includes both the `\input` macro and the inputted content, so the content is included twice.

mdk73:
Your code is working for me out of the box, I get "b", see attached screenshot.
![image](https://user-images.githubusercontent.com/2737768/102193398-55408e80-3ebc-11eb-88f7-2735bc5c23b6.png)
![image](https://user-images.githubusercontent.com/2737768/102193124-f9760580-3ebb-11eb-9719-621512dbafa2.png)

I use a current MikTex- / current plastex-setup (anaconda python distribution) on Win10.

PR #228, opened 2020-09-02
Don't insert \input command after parsing

dalcde:
If we leave the \input command, both the \input command and the input
content are in the source. If this is in the preamble, the imager's
compiler would run the content twice, which results in an error.

See also #179


PatrickMassot:
Thanks!

PR #240, opened 2020-12-16
Use tempfile to name the Imager's temporary LaTeX source file

georgestagg:
This replaces using the filename `image.tex` with a temporary one from `tempfile.mkstemp` when the Imager is used. This avoids a problem with filename collision when a source input file is named `images.tex`.

### DIFF
--- a/plasTeX/Imagers/__init__.py
+++ b/plasTeX/Imagers/__init__.py
@@ -441,6 +441,9 @@ class Imager(object):
     # The compiler command used to compile the LaTeX document
     compiler = 'latex'
 
+    # The filename for temporary LaTeX source
+    tmpFile = Path('images.tex')
+
     # Verification command to determine if the imager is available
     verifications = []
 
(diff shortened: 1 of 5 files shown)


PatrickMassot:
Thanks, this is a nice idea! However the code doesn't quite makes sense. `mypy` says:
```mypy
plasTeX/Imagers/__init__.py:717: error: "str" has no attribute "with_suffix"
plasTeX/Imagers/__init__.py:718: error: "str" has no attribute "with_suffix"
plasTeX/Imagers/pdf2svg.py:18: error: "str" has no attribute "with_suffix"
plasTeX/Imagers/pdf2svg.py:20: error: "str" has no attribute "with_suffix"
plasTeX/Imagers/pdf2svg.py:30: error: "str" has no attribute "with_suffix"
plasTeX/Imagers/dvisvgm.py:16: error: "str" has no attribute "with_suffix"
```
I know I should handle the fact that Travis no longer works and switch to GitHub Actions to run CI (but you are [encouraged](https://github.com/plastex/plastex/blob/master/CONTRIBUTING.md) do it locally before opening PRs).

georgestagg:
Ah, sorry. I think the problem was caused by initialising `tmpFile` as a string,
```
tmpFile = 'images.tex'
```
when it should have been a `Path` object from `pathlib`. I've switched the line and force pushed so that the line reads,
```
tmpFile = Path('images.tex')
```
which seems to make `mypy` happy. I'll try and get a proper testing scheme set up soon, which I'll need for #241 in any case.

PatrickMassot:
Thanks!

PR #241, opened 2020-12-16
Implement xcolor package

georgestagg:
This adds partial support for the `xcolor.sty` LaTeX package, for use with the HTML5 renderer. The main features of `xcolor` are implemented, including defining colour series and colour mixing.

Note that using `\color` to set the colour of a block of content spanning multiple paragraphs does not work correctly. Possibly this is hard to fix due to the use of `<span>` in the renderer template, but I think the same problem also exists in `color.py` in any case.

### DIFF
--- a/plasTeX/Packages/xcolor.py
+++ b/plasTeX/Packages/xcolor.py
@@ -22,9 +22,10 @@ Current Issues:
       color profiles.
 """
 
-from plasTeX import Command, Environment, sourceChildren
-from enum import Enum
-from typing import Optional, Union
+from plasTeX import Command, Environment, sourceChildren, Node
+from enum import Enum, auto
+from typing import Optional, Union, Dict, List
+from typing_extensions import TypedDict
 from re import Scanner # type: ignore
 import math
 
@@ -42,18 +43,18 @@ class ColorModel(Enum):
     specified in. They are not converted until mixing with a color in another
     model is requested, or a HTML representation is required.
     """
-    natural = 0
-    rgb = 1
-    cmy = 2
-    cmyk = 3
-    hsb = 4
-    gray = 5
-    RGB = 6
-    HSB = 7
-    HTML = 8
-    Gray = 9
-    Hsb = 10
-    wave = 11
+    natural = auto()
+    rgb = auto()
+    cmy = auto()
+    cmyk = auto()
+    hsb = auto()
+    gray = auto()
+    RGB = auto()
+    HSB = auto()
+    HTML = auto()
+    Gray = auto()
+    Hsb = auto()
+    wave = auto()
 
 class ColorError(Exception):
     """There was some problem with color definition or mixing."""
@@ -76,10 +77,32 @@ class ColorParser:
     parseColor() method, which returns a single Color object with any requested
     color mixing or other operations applied.
     """
-    tokens = []
-    colors = {}
-    target = ColorModel.natural
-    current_color = None
+    Spec = TypedDict('Spec',{
+        'element': 'ColorParser.Elements',
+        'id': str,
+        'values': List[Union[int,float]]
+    }, total=False)
+
+    ColorMix = TypedDict('ColorMix',{
+        'element': 'ColorParser.Elements',
+        'pct': float,
+        'id': str,
+    }, total=False)
+
+    Token = TypedDict('Token',{
+            'element': 'ColorParser.Elements',
+            'id': str,
+            'ids': List[str],
+            'idlist': List[List[str]],
+            'value': Union[int,float],
+            'values': List[Union[int,float]],
+            'specs': List[Spec],
+            'mix': ColorMix,
+            'mixes': List[ColorMix],
+            'model': ColorModel,
+            'models': List[ColorModel],
+            'color': 'Color',
+    }, total=False)
 
     class Elements(Enum):
         """An enumeration of all syntax elements supported by the parser.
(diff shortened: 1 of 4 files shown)


PatrickMassot:
Thank you very much for this contribution! Could you please add documentation and tests? The current version is basically impossible to maintain (or even review). plasTeX is a very old project. It started before python 2 and many parts of it were written a long time before what are currently seen as basic python writing best practices. Going over the whole code base would be a huge task, but we try to do better with new parts. I'm not the original author and maintainer, so I think we should be nice to whoever will maintain this 10 or 20 years away from now.

The minimal documentation requirement if that we should know the expected argument types and return type of every function. The preferred way to do that is to add [type annotations](https://mypy.readthedocs.io/en/stable/getting_started.html#function-signatures-and-dynamic-vs-static-typing). If you don't know how to do this and don't want to learn, or don't have time, then putting this information in docstrings is fine.

A lot of testing should be easy to write because most of your code has nothing to do with plasTeX. Even if you don't know pytest, you can do a lot without learning anything. You simply need to create a file `unittests/xcolor.py`, import you work there and add functions whose name starts with `test_` and whose body contain `assert` commandes. For instance:
```python
from plasTeX.Packages.xcolor import ColorParser

def test_colorparser_basic_rgb_parsing():
    c = ColorParser().parseColor('0,0,.9', 'rgb')
    assert c.r == 0 
    assert c.g == 0 
    assert c.b == .9
```
(Note: I've never used the real `xcolor` package and it took me an awful lot of time to understand enough of what you code is meant to do to write the above snippet, this illustrate my point about documentation). Then you can install pytest and type `pytest -k xcolor` to run only your tests (the full test suite is pretty slow).

Thanks again for sharing your work. I really hope we'll manage to merge it.

georgestagg:
Yes, I completely understand.

Indeed, most of the code is either named colour definitions or for parsing of complex `xcolor` specifications such as,
```
cmyk:red,1;--foo!50green!25!blue!60!red!![7],11.25;blue,-2>wheel,1,12
```
although admittedly real life use rarely leads to anything this complex.

Testing should not be difficult, I can just check the mixed output colours are as expected. There's plenty of examples in [the xcolor docs](https://anorien.csc.warwick.ac.uk/mirrors/CTAN/macros/latex/contrib/xcolor/xcolor.pdf) I can sample from.

I should hopefully be able to start work on documentation and adding tests at some point in the early new year.

PatrickMassot:
Great! Don't forget to `git pull --rebase` before resuming work on this.

georgestagg:
I've added some documentation and type hints to the package source and added some unit tests. Hopefully that's looking a little easier to maintain now!

PatrickMassot:
Thank you very much, this is already much better! Since you seem ok with type annotations, could you go a bit further in two directions:
* Use more precise annotations for lists and dicts. You can import `Dict` and `List` from `typing` and use annotations like `Dict[str, 'Color']` to annotate a dictionary whose keys are strings and values are `Color`. Similarly you can use say `List[Color]` (or `List['Color']` if `Color` is not yet defined).
* Annotate variable declarations, especially in `__init__` when declaring members or when defining class variables. For instance, when reading [those lines](https://github.com/plastex/plastex/blob/c75f4b4ec6015deca17a9b93337d12eb21898e2c/plasTeX/Packages/xcolor.py#L79-L80) we cannot say what `tokens` and `colors` are meant to hold. You could write `tokens: List[Dict[str, 'Elements']] = []` for instance (I'm only guessing here, maybe this is wrong). By the way, why are those class variables? Unless I'm misunderstanding your code, this seems very wrong, especially with mutable types like list and dicts. Those variables will be shared among all instances.

Is there any reason to use so many lines when defining `Enum`? Do those numbers have an importance or you could simply let python choose them? For instance you could type:
```python
CM = Enum('ColorModel', 'natural rgb cmy cmyk hsb gray RGB HSB HTML '
                                          'Gray Hsb wave')
```
I don't mind, I'm only curious.

I'm puzzled by your function defining dictionaries too. Why repeating the name of the dictionary at every line like this? 
For instance you have a function starting with:
```python
def x11names() -> dict:
    """Return the x11names dictionary.
    The full list of 317 colors traditionally shipped with a X11 installation,
    loaded when xcolor is invoked with the x11names option.
    """
    colors={}
    colors['AntiqueWhite1'] = rgbColor(1,.936,.86)
    colors['AntiqueWhite2'] = rgbColor(.932,.875,.8)
    colors['AntiqueWhite3'] = rgbColor(.804,.752,.69)
    colors['AntiqueWhite4'] = rgbColor(.545,.512,.47)
    ...
```
Why not starting with:
```python
def x11names() -> dict:
    """Return the x11names dictionary.
    The full list of 317 colors traditionally shipped with a X11 installation,
    loaded when xcolor is invoked with the x11names option.
    """
    return {
        'AntiqueWhite1': rgbColor(1,.936,.86),
        'AntiqueWhite2': rgbColor(.932,.875,.8),
        'AntiqueWhite3': rgbColor(.804,.752,.69),
        'AntiqueWhite4': rgbColor(.545,.512,.47)
        ...
```
And why having a function at all by the way? Why not a constant `X11NAMES = { ... }`?

georgestagg:
Thanks for the suggestions! The changes I've made are below:

I've further tightened up the type annotations. I found this a bit tricky to get right as more precise type annotations caused `mypy` to report many extra problems in the static analysis. In particular, the type of `value` in `ColorParser` tokens is highly dependent on the value of `element`, so `mypy` struggled to infer what real types are actually being used.  I was able to tweak the `ColorParser` enough to keep `mypy` happy but it required some fairly complex type annotations using `TypedDict` and splitting the values out into differently named keys. I've never used type annotations before so I suspect might have missed an "easy" method to do this.

I had forgotten about the subtle difference between class and instance variables. I've made changes to remove unneeded class variables and have type annotated the instance variables in ``__init__`` instead.

I didn't know about the functional API for `Enum`. The values don't matter, so I would switch to the nicer looking one liner but I don't know how best to attach a docstring to the enum in that case. Explicitly setting `__doc__` seems strange to me. Since the numbers don't matter I've switched to using `auto()` to make it easier if things are changed later.

I originally generated the colour dictionaries using a script converting the original colour definitions into something Python could understand. I hadn't actually considered rewriting the output until now, but your suggestion is indeed better. I've stuck to using functions, rather than constants. This is because the `basenames` function takes an argument so that the base colours are converted into the document's target colour model. The other functions don't need that, but this way they are all consistent.

I'm glad I wrote those tests. They really helped ensure I wasn't breaking the code while making changes!

PatrickMassot:
I'm sorry about the delay. Type annotations have been painful to write because your coding style is very adverse to typing. I don't want to torture you, especially since you carefully wrote tests. So I'll merge this if you are tired of working on it.

But I strongly believe the same reasons that make your code hard to annotate make it hard to understand, reason about and maintain. Lack of structure mean information is hidden. Typing information contain information about the role and purpose of things.

The specification case is the easiest. They are many dictionaries in this code. Some of them are meant to be color specifications that will be consumed by `Color.makeColor`. How do we convey this information? Some part of your code assume these dictionary will have certain keys. Sometimes this is explicit, because you check there are present, but sometimes you simply assume this in the last branch of an `if` ("if everything else failed then surely this key should be there").

Why not creating a `Spec` class? Going over the board you could even create a `Spec` class and three subclasses depending on how the spec was provided. But let's experiment with only one class. See https://gist.github.com/PatrickMassot/b0fe6077bf31fb7392cf5d039dcf1bf9 (you should probably download it and compare it with your version using a nice diff tool such as [meld](https://meldmerge.org/)). There isn't much difference and the unit tests are trivial to adapt, but I think the intent is clearer, and it includes one `assert` call which clearly states what you are expecting (and make `mypy` happy).

Using the same single class trick probably wouldn't work for `Token`. I mean the code would run fine and mypy would be happy, but the intent wouldn't be clear. Here you probably need subclasses. Then you can clearly distinguish in your annotations between functions with input or output any kind of token and the one that take or output specific kinds of tokens.

PR #243, opened 2021-01-15
some corrections in Ifs.py and some improvements in dealing with \if...\fi

mennucc:
Dear all,

 I submit 4 patches to improve unittest.Ifs and the managing of complex  '\if ..\fi' TeX statements.

Patch 47d5400 adds a warning if there is an unterminated '\if' ; indeed standard pdflatex does warn in this case.

Patch bf388fc  adds a unittest for this warning. There is a standard method to check if the code logs a warning when it should (see [1]) but this fails in plastex, so the unittest is there but is marked with `@pytest.mark.xfail` ; there are notes in the code explaining what is going on; if someone wish, the Logging code in plastex should be changed to be compatible with [1].

 Regards a.

[1] https://docs.python.org/3.4/library/unittest.html#unittest.TestCase.assertLogs


mennucc:
I added 'chmod +x' . The problem is with `gitg` , nice program, but it has this nasty bug, in that it deletes the `x` flag on commits.

mennucc:
I see that coverage has slightly decreased; unfortunately I do not (yet) understand this tool.. sorry. In particular  the tool should show source files, but it is not currently showing them.

dalcde:
On Fri, Jan 15, 2021 at 06:26:49AM -0800, A Mennucc wrote:
>I see that coverage has slightly decreased; unfortunately I do not (yet) understand this tool.. sorry. In particular  the tool should show source files, but it is not currently showing them.

I wouldn't worry too much about code coverage.

PatrickMassot:
Thanks, but I'm a bit confused by what this PR achieves. There is no test that used to fail (and was marked as an expected failure) and now succeeds, right? You added a new test but it's failing (for reasons you explain) and then there is a new successful test `testNewIfIf`. Would that new test fail with current master?

You shouldn't worry about the executable flag stuff. This is a remnant from the distant past where those tests files were run directly. I should probably remove all these.

mennucc:
hi Patrick, 

the patch 3202b5d  does correct an error in master. The error is triggered when you put something like `\newif \iffoobar` inside an `\if` 

the code in master does not recognize the `\newif \iffoobar` but thinks that `\iffoobar` is the start of a nested if, and then it looks for a closing `\fi`

Unfortunately the test `testNewIfIf`is incorrect. Here is a test that works

```
compare_output(r'before  \iftrue always \newif \iffoobar up \else down \fi then \iffoobar one \else two \fi allend')
```

What happens?

`pdflatex` groups the above as follows 
`'before  \iftrue { always \newif \iffoobar up } \else { down } \fi then \iffoobar { one } \else { two } \fi allend'`
and outputs `'before always up then two allend'`

the current master of `plastex` groups as follows
`before  \iftrue { always \newif { \iffoobar { up }  \else { down }  \fi  }then \iffoobar one \else two \fi allend`
and the initial `iftrue` has no matching `fi` (one of my patches will print a warning)
and for reasons I do not properly understand, `plastex` will output   `before always up down then two allend`

mennucc:
hi, I changed the patch and sent you another pull request. I close this. a.

PR #247, opened 2021-02-04
Better

mennucc:
hi, I changed the last patch as discussed. a.

### DIFF
--- a/plasTeX/TeX.py
+++ b/plasTeX/TeX.py
@@ -557,10 +557,15 @@ class TeX(object):
         cases = [[]]
         nesting = 0
         correctly_terminated = False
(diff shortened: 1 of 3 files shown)


mennucc:
this is mostly the same as previous pull request https://github.com/plastex/plastex/pull/243 , but for correction on topmost patch

PatrickMassot:
Thanks for this contribution! I don't understand why the tests don't see logging information. The `plasTeX.Logging` is using the standard logging module, but with custom classes. I don't know what this customization is trying to do here, but I fixed your test in 
563c92b, using the mocking approach that we already use in other tests. I also took the opportunity to convert this file to `pytest` since I needed `pytest` for this fix.

Issue #233, opened 2020-10-14
German translations for babel=ngerman


ugoertz:
It would be nice to have the translations here
https://github.com/plastex/plastex/blob/e1c63556ada2dfa576eea1ffee2747f7f6a0cc58/plasTeX/i18n.xml#L42
also for `babel=ngerman` (German with the "new", i.e., current orthography rules).

PatrickMassot:
Fixed in #250.

PR #250, opened 2021-03-27
Add de-ngerman (current German orthography) to i18n.xml

ugoertz:
See issue #233.


PatrickMassot:
I'll have to trust you here, I don't speak German...

PR #251, opened 2021-03-27
Add amsbook dummy package

ugoertz:
With this "dummy" file, plasTeX also accepts TeX files with documentclass amsbook.

### DIFF
--- a/plasTeX/Packages/amsbook.py
+++ b/plasTeX/Packages/amsbook.py
@@ -1,0 +1,3 @@
+#!/usr/bin/env python
+
+from plasTeX.Packages.book import *


PatrickMassot:
I'm slightly worried that this could cause confusion if authors expect some effect from this class. Of course a lot of effects of the real amsbook class are purely stylistic and wouldn't make sense in a plasTeX context. But are you sure there is nothing else?

ugoertz:
Thanks for the quick reply. I understand the concern, and in fact I am not sure: I have not checked this against all features of the amsbook package. What I can say is that I am using it successfully for half a year now for a moderately large document (the pdf has 350+ pages, the plasTeX is [here](https://math.ug/la1-ws2021/)).

I suppose it is a trade-off between possibly causing trouble for users who use more subtle features of the amsbook document class and on the other hand enabling those who do not, to get started with plasTeX. For me it was one of the small hurdles in the beginning in order to get going, in the phase where I was making a decision between the options to produce HTML from LaTeX, so I though I should share it. (But this also means that for me it is not so important to get this change into the plasTeX repository - now that I have figured it out, it is of course easy to carry this small file along in my own fork.)

Maybe a warning should be added stating that this is only partially tested? If you think so, what would be a good way to do this - as a comment in the file or as a print statement somewhere?

The amsart package in plasTeX is similarly simply an import of the usual article class. Of course, I do not know whether a more thorough check was done in that case.

PatrickMassot:
Ok, I merged and added an information log in 9cd40c4. I don't really want this to be technically a warning, information sounded right.

Issue #232, opened 2020-10-13
Problem with sorting of index items


ugoertz:
Currently the items in the index (produced by printindex) are usually not sorted correctly and therefore are not grouped correctly (i.e., for each starting letter there will be several groups, not just one). The reason is that in commit d8126bd580392976e3adba5240bf8de2e1540b22 there was a change to the following passage in Base/Latex/Index.py in the `__lt__` method of the IndexEntry class. Previously, it was

```python
    result = cmp(...)
    if result == 0 and len(self.key) != len(other.key):
        return cmp(len(self.key), len(other.key))
```
where result being 0 indicated that the cmp in the first line resulted in an equality. In the new version the cmp was replaced (prompted by a change to the way the Python comparison methods work, if I understand correctly) by a < comparison and the middle line of the above excerpt is now

https://github.com/plastex/plastex/blob/e1c63556ada2dfa576eea1ffee2747f7f6a0cc58/plasTeX/Base/LaTeX/Index.py#L420

However, now the check for `not result` does not check for equality in the previous < comparison, but just means that we had >= in that comparison. So even in the > cases, the outcome could be changed, and therefore the sorting is messed up.

The `not result` should be replaced by a check whether the two items that are compared in the previous line are actually equal.

Thanks for your work!

PatrickMassot:
Thank you very much for investigating and posting this message! What would help even more is an example where sorting fails. I totally believe you see issues, but I wasn't able to reproduce (you can see [here](https://www.imo.universite-paris-saclay.fr/~pmassot/enseignement/mat553/sect0037.html) a non-trivial index that seems perfectly fine, even including non-ascii sorting). With an example I'd be able to write a test and fix the code.

ugoertz:
Thanks for your answer. Maybe it works in your case because for most terms there is only one entry? I am not sure.

Anyway, I append an example TeX file and a screenshot of the index which I get.
[example.txt](https://github.com/plastex/plastex/files/5396738/example.txt)
![Screenshot from 2020-10-17 23-34-11](https://user-images.githubusercontent.com/668977/96354006-cdafec80-10d1-11eb-9464-e36233dff4af.png)

For me, things work fine after replacing the `__lt__` method by the following:

```python
    def __lt__(self, other):
        key_self = list(zip([collator(x) for x in self.sortkey if isinstance(x, str)],  
                            [collator(x.textContent) for x in self.key],  
                            self.key))  
[...]

PR #252, opened 2021-03-27
Fix sorting of index entries

ugoertz:
This is a follow-up to issue #232 which fixed one problem with indexing, but left another one open, as I noticed later. See the new test which fails without the proposed change to DOM/__init__.py.

### DIFF
--- a/plasTeX/DOM/__init__.py
+++ b/plasTeX/DOM/__init__.py
@@ -1176,9 +1176,9 @@ class Node(object):
             if res: return res
             if self.hasChildNodes() and other.hasChildNodes():
                 return self.childNodes < other.childNodes
-        except AttributeError:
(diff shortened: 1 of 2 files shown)


PatrickMassot:
Thanks! Could you please use in the test a TeX source that pdflatex + makeindex can actually use, so that we can compare? (We don't have to put the comparison in the test, but I'd like to see it before merging).

PatrickMassot:
I'm sorry about CI integration complaining about things unrelated to your PR. You can rebase on master to fix that.

ugoertz:
Thanks, yes that makes sense. I expanded the TeX so that it can be used with pdflatex/makeindex.

PatrickMassot:
Thanks!

PR #255, opened 2021-03-28
Add placeholder for pgfplots package

ugoertz:
Adding a placeholder file allows using this package in tikzpictures.
(It does cause some plasTeX warnings, as TikZ itself does - I am not sure what the policy/the right approach about this is. In a sense I do not understand why plasTeX should complain about things inside tikzpicture environments which it does not understand (such as node, draw, etc.), since they will be dealt with separately anyway. Or is this a problem with my setup? Anyway, I think that if one wants to address this then one should find a solution that takes the existing tikz package into account, as well.)


PatrickMassot:
> I do not understand why plasTeX should complain about things inside tikzpicture environments which it does not understand (such as node, draw, etc.), since they will be dealt with separately anyway. 

It doesn't complain actually. The issue is that your empty package doesn't load tikz, so plasTeX has no idea what `tikzpicture` means. I tried to implement a more correct version at 6663816dc31543242. Could you confirm this solves your issue?

ugoertz:
Thanks for improving my simplistic/naive submission. Also, thanks for the further comments - after some further investigation I found that the warnings are not caused by the pgfplots package, but in a different way (whose cause is very likely that I do not understand enough about how plasTeX works, as well). Sorry for the "false alarm".

(To explain - but feel free to ignore this: I have a command twominipages which I can use as
```latex
\twominipages{6cm}{6cm}{First column}{Second column}
```
and which for pdflatex is a macro which results in two minipages of the given widths and with the given content. For plasTeX I have defined a command as
```python
class twominipages(Command):
    args = 'dim1:float dim2:float text1 text2'
```
(plus an appropriate HTML template). I can put a tikzpicture as the content of one of the minipages (in the sense that this "works" with pdflatex and plasTeX), but I get warnings from plasTeX (such as `WARNING: unrecognized command/environment: draw`).)

Anyway, as far as I can see, the pgfplots.ps Package that you added works well.

PatrickMassot:
I'm closing this since the original issue is solved, but don't hesitate to open another issue with your `twominipages` problem, including your full python code and an example tex fileso that I can reproduce the problem.

PR #264, opened 2021-07-02
fix double dollar construction rendering

efiminem:
Previously, the `$math$$math$`-type construction (with two dollar signs in a sequence without spaces) was treated incorrectly, because it was rendered as if it was a beginning of `displaymath` context inside. It led to a serious problem, because the whole document was within incorrect context after that moment.

### DIFF
--- a/plasTeX/Base/TeX/Primitives.py
+++ b/plasTeX/Base/TeX/Primitives.py
@@ -78,7 +78,11 @@ class MathShift(Command):
         current = self.ownerDocument.createElement('math')
         for t in tex.itertokens():
             if t.catcode == Token.CC_MATHSHIFT:
-                current = self.ownerDocument.createElement('displaymath')
+                # Case where we have $   $$    $ construction
+                if inEnv and inEnv[-1] is not None and type(inEnv[-1]) is type(current):
+                    tex.pushToken(t)
+                else:
+                    current = self.ownerDocument.createElement('displaymath')
             else:
                 tex.pushToken(t)
             break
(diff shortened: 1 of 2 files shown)


PatrickMassot:
Thank you very much for this contribution. Encouraging people to write such TeX feels bad, but TeX itself does it...

I have two requests before merging. First you should rebase on master (I just push two commits to fix the mess created by a mypy super backward incompatible change). More importantly, I'd like to have tests for this commit. It's even better if they go in the same commit as the fix so they act as commit documentation.

You can edit `unittests/Primitives.py`. Add `from plasTeX.TeX import TeX` at the very top of the file and then, at the very bottom:
```python
def test_evil_double_dollars():
    tex = TeX()
    tex.input('$a$$b$')
    nodes = tex.parse().getElementsByTagName('math')
    assert len(nodes) == 2
    assert nodes[0].textContent == 'a'
    assert nodes[1].textContent == 'b'

def test_more_evil_double_dollars():
    tex = TeX()
    tex.input('$a$$$b$$$c$')
    doc = tex.parse()
    mathnodes = doc.getElementsByTagName('math')
    assert len(mathnodes) == 2
    assert mathnodes[0].textContent == 'a'
    assert mathnodes[1].textContent == 'c'
    displaymathnodes = doc.getElementsByTagName('displaymath')
    assert len(displaymathnodes) == 1
    assert displaymathnodes[0].textContent == 'b'
```
and maybe other variations.

efiminem:
I've updated the request with the tests

PatrickMassot:
Thanks a lot!

Issue #263, opened 2021-06-15
Bibliography with natbib problem


samtsevich:
Once again I write you to describe the problem with natbib usage in TeX while converting to the html.
Below you will find 2 file that represent the problem.

If you will try to run:
`plaster tmp.tex`
you will find this error.

Hope that you can fix that easily.

UPD. I did provide a wrong archive with necessary files. Now, you can find proper one below of this comment.
[Archive.zip](https://github.com/plastex/plastex/files/6676051/Archive.zip)

efiminem:
This issue is fixed in #265

PatrickMassot:
I don't see how this test file with one bibliographic reference can be a test for a bibliography **sorting** function. I already told you [one year and half ago](https://github.com/plastex/plastex/issues/234#issuecomment-750855428) that modifying this function so that it compiles was trivial. But I didn't want to merge code that I have no idea what it is meant to do and whether it does it. And I don't want to spend time trying to understand what a super obsolete LaTeX package is meant to do.

Issue #234, opened 2020-11-11
Bibliography problem


samtsevich:
Previously, I had a plastex built on the python2.7. But I made `git pull`, and I had to reinstall it, basing on the python3.8, which looks fine.
So, I was trying to run the following command:
`plastex file.tex`

And got a very long log with a very strange error related to the citation part:
```
plasTeX version 2.1
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/article.py
 )
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/graphicx.py
 )
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/color.py
 )
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/url.py
 )..
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/multicol.py
 )..
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/natbib.py
 )..
WARNING: No Python version of bibentry.sty was found
WARNING: Will now try to load a LaTeX implementation of bibentry.sty
 ( /usr/local/texlive/2020/texmf-dist/tex/latex/natbib/bibentry.sty )
WARNING: unrecognized command/environment: @ifstar
...........................................................................
.........................................................................
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/tocbibind.py
 )..
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/amsmath.py
 )
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/hyperref.py
 )..
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/listings.py
 )..................
 (loading package /usr/local/lib/python3.9/site-
   packages/plasTeX/Packages/fancyhdr.py
 )..... ( ./title_page.tex )..................... ( ./1_history_USPEX.tex )
...........................................................................
...........................................................................
....................... ( ./2_Getting_started.tex )........................
...........................................................................
........................... ( ./3_input_and_output_files.tex ).............
...........................................................................
.....................................................
 ( ./4_input_options.tex ).................................................
...........................................................................
...........................................................................
...........................................................................
...........................................................................
...........................................................................
...........................................................................
...........................................................................
...........................................................................
...........................................................................
...........................................................................
...........................................................................
[...]

PR #265, opened 2021-07-02
fix sort in natbib

efiminem:
Previously, a sorting function in natbib was not working because it requires a sorting key instead of comparator.

### DIFF
--- a/plasTeX/Packages/natbib.py
+++ b/plasTeX/Packages/natbib.py
@@ -222,11 +222,12 @@ class NatBibCite(Base.cite):
         for key in self.attributes['bibkeys']:
             if key in b:
                 items.append(b[key])
-        if bibpunct.punctuation['style'] in 'ns' and \
-           ('sort' in opts or 'sort&compress' in opts or 'sortandcompress' in opts):
-            items.sort(lambda x, y: int(x.ref) - int(y.ref))
-            if 'sort&compress' in opts or 'sortandcompress' in opts:
-                items = self.compressRange(items)
+        if len(items) > 1:
+            if self.isNumeric() and \
+            ('sort' in opts or 'sort&compress' in opts or 'sortandcompress' in opts):
+                items.sort(key=lambda x: int(x.ref))
+                if 'sort&compress' in opts or 'sortandcompress' in opts:
+                    items = self.compressRange(items)
         return items
         
     def compressRange(self, items):
--- a/plasTeX/Packages/natbib.py
+++ b/plasTeX/Packages/natbib.py
@@ -453,7 +453,8 @@ class citet(NatBibCite):
         res = self.ownerDocument.createDocumentFragment()
         i = 0
         sameauthor = prevauthor = None
-        for i, item in enumerate(self.bibitems):
+        bibitems = self.bibitems
+        for i, item in enumerate(bibitems):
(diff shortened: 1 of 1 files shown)


PatrickMassot:
Thanks for this contribution. As I explained in https://github.com/plastex/plastex/issues/234#issuecomment-750855428, it would be much nicer to also have a test showing this function does what it is expected to do. And I still have no idea what it is expected to do (and still no idea why people are using natbib in 2021).

efiminem:
Thanks for the answer. I see that you've basically suggested the same fix, haven't noticed it because the issue was closed. I was not diving into details of how the function works, but this particular line of code is obviously not compatible with the newer python and should be fixed (regardless of the purpose of the code). Investigating the package further could be a part of another request. As of using natbib in 2021 - well, sometimes you just have a legacy code.

PatrickMassot:
I don't particularly enjoy being the grumpy old man reluctant to merge code that is obviously an improvement. The point is the original code was written probably more than 15 years ago and nobody took the trouble to write a test at that time. I guess that code was "obviously" correct at this time. Then python3 came and that code broke, and nobody noticed because there was no test (and no natbib user). Fixing this today without writing a test is simply setting up a trap for our successor 15 years away from now.

efiminem:
Ok, I've looked into the code more closely. This function is called when sort option of natbib is used and citations are numbers. Its purpose is to sort the citations which go in one place (e.g. [3, 7, 8, 2] should be changed to [2, 3, 7, 8]). So, in order to see any difference due to this implementation an example should contain such one-place citations.

Note, that due to bad implementation the sorting was also called when list of one-place citations is empty or contained only one element. This is the reason why example provided in #263 was also triggering the sorting despite consisting only of one reference. In addition to converting the method to newer python I've also added two changes:

1.  Now the length of one-place citations is checked and sorting is performed when it contains at least 2 elements.
2. `bibpunct.punctuation['style'] in 'ns'` is changed to `self.isNumeric()` to make it more clear what this part of the code do.

I've also noticed another related problem while following the code. In some places `self.bibitems` was unnecessary called resulting in additional sorting for each already sorted item. I've fixed it in another commit.

As for the tests, I'm not sure what is the best place to put them. Is it `unittests/Packages/sources/`? The simple example which uses the sorting functionality is attached: [sorting.zip](https://github.com/plastex/plastex/files/6759053/sorting.zip).

PatrickMassot:
Sorry about the delay. `unittests/Packages/sources/` would be the place to put a tex file if you want to do a functional end-to-end test, from the tex file to the rendered html. This is not the relevant level here. We want tests testing that sorting option, not the renderer. I started to write tests but I realized this package is written in a way that does not allow handling several files without re-initializing everything. There is a global variable `PackageOptions` which stores the configuration and, much worse, the global class `bibpunct` is modified when reading the options. So writing a series of tests with various options for `natbib` is currently very complicated (tests would have weird interactions with each other). So I will merge this PR, and hope I'll find courage to sanitize this package and write a couple of tests.

PatrickMassot:
@efiminem if you are curious, you can have a look at 4ce3bae0871dd47084050d7038991729e0c6bde3 and fd5ef4270dc3da8be15917204c5eb6257424f6a9.

Issue #277, opened 2022-03-06
text width can become negative


kno10:
Line width can become negative here:
https://github.com/plastex/plastex/blob/8dffc0281fe7b0a0ba6cc5498f597d0492f483cb/plasTeX/Renderers/Text/__init__.py#L144-L166

PatrickMassot:
Thank you for this report. I don't think anyone has used this renderer in the last ten years and I've never looked at this part of plasTeX. I have very limited time for plasTeX and I don't see how the text renderer could become a coding priority but I'll happily consider pull requests improving it, especially if they come with unit tests (or at least minimal examples showing what happens before and after the improvement that I can easily turn into tests).

PatrickMassot:
I'm closing this since there is no activity. What I wrote on March 7th about pull requests still holds of course.

PR #282, opened 2022-03-07
remove extra argument self in TextRenderer.do_footnotetext

kno10:
self is already on the left.


PatrickMassot:
Thanks! This was probably a mistake introduced by the python2 to python3 automatic conversion (and it confirms what I wrote in #277 about having no user for this renderer...).

PR #283, opened 2022-03-07
Fix subsitution logic

kno10:
The replacement pattern is compiled to resolve backreferences, and this may fail if there are backslashes etc. included.
A simple string operation supposedly does what was intended?

```
    s = block_re.sub('%s%s' % (before, block), s, 1)
  File "/usr/lib/python3.9/re.py", line 327, in _subx
    template = _compile_repl(template, pattern)
  File "/usr/lib/python3.9/re.py", line 318, in _compile_repl
    return sre_parse.parse_template(repl, pattern)
  File "/usr/lib/python3.9/sre_parse.py", line 1042, in parse_template
    raise s.error('bad escape %s' % this, len(this))
re.error: bad escape \m at position 52 (line 3, column 51)
```

### DIFF
--- a/plasTeX/Renderers/Text/__init__.py
+++ b/plasTeX/Renderers/Text/__init__.py
@@ -77,7 +77,7 @@ class TextRenderer(BaseRenderer):
             block = self.blocks[int(m.group(2))]
             block = space + block.replace('\n', u'\n%s' % space)
 
-            s = block_re.sub('%s%s' % (before, block), s, 1)
+            s = s[:m.start()] + before + block + s[m.end():]
 
         # Clean up newlines
         return re.sub(r'\s*\n\s*\n(\s*\n)+', r'\n\n\n', s)


PatrickMassot:
Could you please include a unit test that fails with current master and works with your patch (or at least a mwe showing the issue)?

kno10:
Sorry, I don't have a unit test or mwe ready for this - I don't understand the hack that is being done her.
Nevertheless, it's clear that the substitution string should not be parsed for regexp references and similar escape sequences, right?

kno10:
Clearly, `'%s%s' % (before, block)` is an inefficient way of writing `before + block`.
But these need to be escaped, so you'd want `re.escape(before + block)`.
Or you do the entire trivial replacement string concatenation as in my patch: `s[:m.start()] + before + block + s[m.end():]` which avoids escaping and parsing the replacement string altogether.

PatrickMassot:
Let's stop hoping someone will find time to understand why this line was written that way 15 years ago. As far as I know, you are the only user of that text renderer, so let's merge this and see if someone complains. I'm sorry about the delay.

PR #284, opened 2022-03-07
Fix unicode symbols for medium/thick spaces

kno10:
c.f.
https://www.fileformat.info/info/unicode/char/205f/index.htm
https://www.fileformat.info/info/unicode/char/2005/index.htm


PatrickMassot:
Thanks! I'd be curious to know how these strange values came there...

PR #286, opened 2022-03-08
Avoid index out of bounds exception in expandDef

kno10:
This can happen with beamer commands, because the `<>` overlay option is added as additional parameter by beamer.

Just pretend its None then rather than failing.

### DIFF
--- a/plasTeX/__init__.py
+++ b/plasTeX/__init__.py
@@ -1105,7 +1105,7 @@ def expandDef(definition, params):
                 if t.catcode == Token.CC_PARAMETER:
                     output.append(t)
                 else:
-                    if params[int(t)] is not None:
+                    if int(t) < len(params) and params[int(t)] is not None:
                         # This is a pretty bad hack, but `ifx' commands
                         # need an argument to also be a token.  So we
                         # wrap them in a group here and let the


PatrickMassot:
Could you please provide an example showing the issue (here or, better, in a new test) and add a comment in your patch above that modification?

kno10:
Here is a MWE that works fine with pdflatex:
```tex
\documentclass{beamer}
\renewcommand<>{\highlight}[2][structure]{{\textcolor#3{#1}{#2}}}
\begin{document}
\highlight{text}
\end{document}
```
while plastex produces
```
plasTeX version 2.1
 ( /usr/lib/python3/dist-packages/plasTeX/Packages/beamer.py )
ERROR: Error while expanding "highlight" in ./mwe.tex on line 4 (list index
   out of range)
ERROR: An error occurred while building the document object in ./mwe.tex on
   line 4 (list index out of range)
Traceback (most recent call last):
  File "/usr/bin/plastex", line 132, in <module>
    main(sys.argv)
  File "/usr/bin/plastex", line 76, in main
    tex.parse()
  File "/usr/lib/python3/dist-packages/plasTeX/TeX.py", line 440, in parse
    item.digest(tokens)
  File "/usr/lib/python3/dist-packages/plasTeX/__init__.py", line 889, in digest
    for item in tokens:
  File "/usr/lib/python3/dist-packages/plasTeX/TeX.py", line 46, in __next__
    return self._next()
           ^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/plasTeX/TeX.py", line 345, in __iter__
    tokens = obj.invoke(self)
             ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/plasTeX/__init__.py", line 1069, in invoke
    return output + expandDef(self.definition, params)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/plasTeX/__init__.py", line 1017, in expandDef
    if params[int(t)] is not None:
       ~~~~~~^^^^^^^^
IndexError: list index out of range
```

PatrickMassot:
Thanks. I merged that 72926603c44b331fc6eb417d1bd68816b6cee22d taking the opportunity to simplify code around there, and also adding a test.

PR #296, opened 2022-09-14
bug fix for #274 line numbering inconsistency

### LINKED ISSUE #274: Line numbering inconsistency?
The current code appears to sometimes overcount the line numbers.
Right now, I have an error reported as
ERROR: Error while expanding "only" in text.tex on line 197 (string index out of range)
but the file only has 193 lines.
I believe this may be due to verbatim sections.
Sorry, I don't have a MWE yet.

kno10:
When pushing back a newline to be parsed again, decrement the line count. Otherwise the lines get double counted, and reported line numbers are wrong.

Closes #274, MWE there

### DIFF
--- a/plasTeX/Tokenizer.py
+++ b/plasTeX/Tokenizer.py
@@ -302,6 +302,8 @@ class Tokenizer(object):
 
         """
         self._charBuffer.insert(0, char)
+        if char == '\n':
+            self.lineNumber -= 1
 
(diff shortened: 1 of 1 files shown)


PatrickMassot:
Thanks for this fix! I added a test at 4b3290b3db0f8659d71b8afef8dd3cc052a632db.

Issue #164, opened 2020-04-14
Improve documentation for beginners


PatrickMassot:
The documentation is currently a bit outdated, and it will become worse after #162 and #159 if we are not careful. http://plastex.github.io/plastex/tutorials/ needs some love. But the main README, http://plastex.github.io/plastex/ and http://plastex.github.io/plastex/plastex/index.html also need to be updated.

We really need to have several layers of documentation, from people who simply want to convert a simple document and don't really care about tweaking the output to what Pieter would have needed when starting the Gerby renderer. Intermediate levels include people who want to easily change some CSS or change a couple of templates.

grunweg:
@PatrickMassot I just opened a [PR](#301) against the README, with some documentation changes that would've made things easier for me. Perhaps this helps.

PatrickMassot:
There is now an updated getting started guide and reference manual.

grunweg:
Wow, the new documentation looks great. Thanks a lot!

PR #301, opened 2022-10-27
README: add sections on installation, basic usage and how to contribute.

grunweg:
Hi! I just looked at plastex in detail; thanks a lot for this project!
I found getting to grips with it slightly difficult; part is that I didn't find the docs (which are helpful!) until later. I gather this is a [known problem](https://github.com/plastex/plastex/issues/164). Find below some improvements to the README which would have helped me.

Notes.
- The link `https://plastex.github.io/plastex/#users` doesn't exist yet (but doesn't go stale either).
- I would find a section "contributing" helpful (see below). What to put there depends on you :-) I've put some questions/possible options. *Please* let me know if any of these is correct/what else would be accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,46 @@
 # plastex
 
-plasTeX is a LaTeX compiler written in python.
+plasTeX is a LaTeX compiler written in Python.
 Read more at the github page for the plasTeX project:
 http://plastex.github.io/plastex/
 
+## Installation
+plasTeX can be installed like any usual Python package.
+
+For instance, you can install it using [pip](https://pip.pypa.io/en/stable/) by typing
+`pip install plastex`
+at a shell prompt. This will install the latest version of plasTeX packaged [on PyPi](https://pypi.org/project/plasTeX/).
+As of October 2022, the last version is almost three years old: you may want to install the latest version from github. For this,
+- clone the repository, e.g. `git clone https://github.com/plastex/plastex`
+- install the latest version of plasTeX: `pip install ./plastex`
+(Mercurial users using [`hg-git`](https://hg-git.github.io/) can run `hg clone https://github.com/plastex/plastex` in the first step instead.)
+
+Updating an existing version of plastex can be done by updating the github repository, then running `pip install --upgrade plastex`.
+
+## Using plastex
+plasTeX can do many things:
+- it can simply compile your LaTeX document into nice-looking HTML,
+- there are many options to configure the output,
+- plasTeX can also be used as a library from other Python programs,
+- finally, plasTeX can be extended in many ways!
+
+If you just want to compile your document, type `plastex myfile.tex` into a shell and you're good.
+If you're interested in the other options, take a look at the [user guide](https://plastex.github.io/plastex/plastex/index.html).
+If you want to extend plasTeX to match your needs, the [tutorial](https://plastex.github.io/plastex/tutorials/) could be a useful starting point. Or take a look at the different users of plastex](https://plastex.github.io/plastex/#users).
+
+## Contributing
+- Do you welcome contributions in general?
+- Anything particular to watch out for?
+- Are there any "good first issues"? How could one help? Are there any tasks that just need to be done, which one could do?
+If there's something to say on either of they, this could be a place to mention them.
+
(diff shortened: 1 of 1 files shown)


grunweg:
**Why am I making these changes? Where am I coming from?**
*Consider this a mix of various bits of feedback with a rationale for my changes. Skip this if you wish.*

1. How to install: I came here via the [sphere eversion blueprint](https://github.com/leanprover-community/sphere-eversion) whose readme recommends cloning this github repository. That seemed unusual to me (I'm more used to a simple `pip install packagename`), hence I was wondering if the package was also on PyPI (or why not)... Turns out it was, but mentioning that would answer this question :-)

2. plasTeX options
After installation, plasTeX compiled the sample document fine, but I wanted to merge all files in one. The webpage said this was possible - but I didn't see any hints how. The README doesn't have it, so I looked at the `plastex --help` output and went from there. I only saw the [documentation](https://plastex.github.io/plastex/plastex/plastex.pdf) the next day. Linking that more prominently would have helped a lot

3. The github page was not helping me find the documentation. I opened a separate issue about this.

4. The tutorial is very hard to discover. I only found out about this by browsing the github issues.
In my opinion, that's a rather high bar. Linking it more prominently would reduce that.
If the tutorial is somewhat outdated, I'd rather link the outdated ressource than none.
(In my case, figuring out how to extend plasTeX to match my need would be *much* harder without any tutorial. Knowing the high-level structure of things certainly seems very helpful for this. This matches [other recommendations I've seen](https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html))

Relatedly, the post above would even recommend an ARCHITECTURE file in the repository.
If that's a widely followed convention (I don't know), even including a dummy file (with a mere "refer to [this tutorial](link here) for details) would help. Otherwise, either seems equally good to me.

5. Perhaps I'm too used to open source projects that are very [actively](https://brson.github.io/2017/04/05/minimally-nice-maintainer#be-clear-about-what-you-expect) [maintained](https://manishearth.github.io/blog/2016/01/03/making-your-open-source-project-newcomer-friendly/): for me, seeing an explicit "contributions are welcome, this and that is the most useful to us" goes a long way to feel my efforts welcome. Even "this is a side project to all of us, we welcome your patches, but responding might take a while" sets clear expectations and helps.

If I feel contributions were not useful, I might move elsewhere, which seems a lose-lose situation. (I want to scratch an itch here!)

**My perspective/background** so you can calibrate my feedback accordingly: I'm mostly a simple user, but would like to tweak plasTeX in one particular way. I started looking at it yesterday, hence am still quite a beginner to plasTeX.

grunweg:
**Update.** I see that the tutorial used to be mentioned on the main page. Hm. I'd rather link to an outdated tutorial than to none. Perhaps linking it on the README only could be a compromise?

PatrickMassot:
Thank you very much for your contribution. I don't think it really improves the situation however. The idea of the current README is to simply redirect readers to the main webpage of the project which is all about documentation. I will edit this main webpage so that the link to the full documentation comes earlier (although it should definitely be read *after* the main webpage).

I don't understand what you refer to with ` https://plastex.github.io/plastex/#users`. Where did you see that link?

We already have information about contribution located in `CONTRIBUTING.md` which is the standard place to look at. After seeing that suggestion, I just added some information into that file. Thanks for the suggestion.

The tutorial you found is so outdated that nothing good can come of using it. I'll try to find how to remove it.

PatrickMassot:
I should add that I also have a new work in progress tutorial somewhere that I should finish one day. Time is very scarce but I will try.

grunweg:
Thanks for the quick response!
- CONTRIBUTING.md is indeed a much better place for these things. (I find the history remark really helpful.)
- FWIW, I originally didn't look for CONTRIBUTING.md because the README was so short and the webpage didn't mention it (I simply assumed it wasn't there). That's clearly a flawed heuristic --- pointing to it from the webpage addresses this well.
- About the tutorial: sure, makes sense. I'm looking forward to the new tutorial, at some point. (Though it seems *I* will not need it for the time being.)

grunweg:
Actually, brief installation instructions would be useful. Suggested in #303.

PR #304, opened 2022-11-01
Fix a few typos.

grunweg:
I found another typo; here you are. Thanks for all your work on this project!

(As a side note, the webpage has a typo, "developpment". Neither very important nor urgent.)


PatrickMassot:
Thanks! The filename thing was clearly intentional but I don't see a reason to keep it.

PR #306, opened 2022-11-16
Update INSTALL

profmel:
Python 3.5 does not exist. The most recent stable release is version 3.11


PatrickMassot:
This is not how software numbering works. The `3.11` in `python 3.11` is not a decimal number, it is a dot-separated list of natural numbers. See e.g. https://en.wikipedia.org/wiki/Software_versioning for more information.

PR #307, opened 2023-01-02
CONTRIBUTING.md: remove mention of version 3

grunweg:
Version 3 has been released to PyPi now, so I guess this can be dropped :-). Many thanks for all your work maintaining this project!


PatrickMassot:
Thanks for this nice catch and nice words!

PR #314, opened 2023-02-07
Fix overescaped backslash in tabbing

kno10:
An additional `\` is automatically prefixed, c.f., `EndRow` in other contexts.

### DIFF
--- a/plasTeX/Base/LaTeX/Tabbing.py
+++ b/plasTeX/Base/LaTeX/Tabbing.py
@@ -16,7 +16,7 @@ class tabbing(Environment):
         macroName = '>'
 
     class EndRow(Command):
-        macroName = '\\\\'
+        macroName = '\\'
    
     class kill(Command):
         pass


PatrickMassot:
Could you please provide a test or at least an example where this patch improves things?

kno10:
This causes doubles backslashes. The macroName will already be prefixed by an `\` elsewhere.
See all the other `macroName` in this file, such as line 16, which is the latex command `\>` but only contains `'>'`.

See also the following working instances of `EndRow`:
https://github.com/plastex/plastex/blob/6182161533d6075bee033fcdcdbefd0f8b8d8997/plasTeX/Base/LaTeX/Arrays.py#L91-L93

https://github.com/plastex/plastex/blob/6182161533d6075bee033fcdcdbefd0f8b8d8997/plasTeX/Packages/xy.py#L9-L11

kno10:
Also, it is missing `args = '* [ len:dimen ]'` to allow modified break lengths, as in the regular `\\` command.
https://github.com/plastex/plastex/blob/72926603c44b331fc6eb417d1bd68816b6cee22d/plasTeX/Base/LaTeX/Breaking.py#L23-L25

PR #318, opened 2023-02-12
split level is ignored if template does not contain a space?

kno10:
This works with the default template `index [$id,sect$num(4)]` *but* it will disable splitting for the template `sect$num(4)`? IMHO, the dollar is more indicative of a template that can generate suitable file names?


PatrickMassot:
Thanks for your contribution but this is not preserving the specification from the documentation which allows giving a static list of file names. 

Also, what do you hope to achieve `sect$num(4)` template? Did you test your code with this filename template and a document containing several sections?

kno10:
Some sections might not have an id, hence the `sect$num(4)` fallback.

PR #319, opened 2023-02-12
remove unused `renderMethod`

kno10:
Apparently all renderers work via to string conversion? If ever needed, this is easily added back, but its your decision whether to keep this.


PatrickMassot:
I've never seen that and it does not appear in the documentation, so let's remove it indeed.

PR #327, opened 2023-02-23
Advance adances a token/counter, not a number

### LINKED ISSUE #322: Parsing of \advance
The `\advance` command is currently implemented as follows:
https://github.com/plastex/plastex/blob/aa5de5c85081e0f352716867ce62943594aa2a9a/plasTeX/Base/TeX/Primitives.py#L466-L470

In my opinion this is incorrect, as you do not advance a number, but a length, i.e., it is supposed to be something like `\advance<name> by <number><unit>`.

This usually would lead to lost tokens (the `<name>` is lost) when parsing code that contained `\advance`, e.g., the newellipse from https://tex.stackexchange.com/a/24621

When I removed the entire `invoke` and made the command a `pass`, things appear to work much better.

kno10:
I believe this is the correct fix for Closes #322

### DIFF
--- a/plasTeX/Base/TeX/Primitives.py
+++ b/plasTeX/Base/TeX/Primitives.py
@@ -465,7 +465,7 @@ class active(CountCommand):
 
 class advance(Command):
     def invoke(self, tex):
-        tex.readArgument(type='Number')
+        tex.readArgument(type='Tok')
         tex.readKeyword(['by'])
         tex.readArgument(type='Number')
 


PatrickMassot:
Thanks, but this is still not correct. The argument after `by` is indeed sometimes a dimension but it can also be a number when the name refers to a counter. In addition the current implementation is clearly a stub, it reads arguments but does nothing with them. What do you expect from your code exactly?

kno10:
See #322

PR #330, opened 2023-02-24
Fix footnote support in beamer

kno10:
The syntax is not `< overlay > [ options:dict ] [ num:int ] self`, but the num parameter is *replaced* to allow the `frame` keyword. This is also possible for the footnotetext; footnotemark is more like regular text and does not appear to support overlay specifications?


PatrickMassot:
I don't really understand the issue but I'll merge this since you are very probably the only user of plasTeX that tries to convert beamer files.

PR #332, opened 2023-02-26
doc: pushToken does not pop from output

kno10:


PatrickMassot:
Indeed it seems this docstring should have been updated by https://github.com/plastex/plastex/commit/279e47fe0daeddbe196ed4531c1441a7afb9bf33 which removed the poping.

PR #333, opened 2023-02-26
Avoid unnecessary list copies/reversal

kno10:
Also, adjust docstring escaping (NewMacro docstring was incorrect), prefer raw strings.

### DIFF
--- a/plasTeX/Tokenizer.py
+++ b/plasTeX/Tokenizer.py
@@ -325,9 +325,7 @@ class Tokenizer(object):
 
         """
         if tokens:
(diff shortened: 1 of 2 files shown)


PatrickMassot:
Thanks. All the stupid `list(stuff.keys()):` come from Tim's port to python3. I thought I removed all of them, but keep finding new ones from time to time.

The other more substantial changes do not seem to work, there are failing tests. Could you rebase on master and try to investigate? Otherwise I will implement a more conservative version of this PR.

PR #334, opened 2023-03-03
Add a log/debug function for jinja2 templates

kno10:
This allows debug logging in jinja2 templates via `{{ "string" |debug }}` or `{{ "string" |log }}` (currently, both do the same).

### DIFF
--- a/plasTeX/Renderers/PageTemplate/__init__.py
+++ b/plasTeX/Renderers/PageTemplate/__init__.py
@@ -41,9 +41,15 @@ else:
         print("\nYou can inspect obj and config.\n")
         pdb.set_trace()
 
+    def _jinja2_log(t):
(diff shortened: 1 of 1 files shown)


PatrickMassot:
Thanks for your contribution. But this cannot be used by other people if you don't write any documentation. 

Also, do you have any use case that is not covered by the existing `{{ debug() }}` documented [here](http://plastex.github.io/plastex/plastex/sect0048.html)?

kno10:
It has been a long time, but probably one motivation was to log certain values *after* some jinja2 processing, dropping into an interactive debugger will help with inspecting the input to the jinja2 template, but how to access what jinja2 produces when building larger templates?

PatrickMassot:
Ok, so I'm willing to merge this if you include documentation.

PR #336, opened 2023-03-08
Avoid creating paragraphs in math matrixes

kno10:
The current solution does not work, because it does not recursively pass the `par` parameter, and its cludgy to add this because `source` is a property.
Because of this, plastex inserts newlines into arrays, e.g.,
```tex
\documentclass{article}
\usepackage{amsmath}
\begin{document}
\begin{align*}
\left(\begin{matrix}-1 & -2 \\ -1 & -2 \end{matrix}\right)
\end{align*}
\end{document}
```
renders the equation as
```html
<div class="displaymath" id="a0000000006">\begin{align*} 
\left(\begin{matrix} -1 

&  -2 

\\ -1 

&  -2 

\end{matrix}\right)
\end{align*}</div>
```
with the double newlines being added because of the paragraphs in each array cell.

This patch does not create paragraphs when in math mode. This allows removing the `par=` parameter, and some places handling this special case, e.g., in `sourceChildren`. An alternate approach would probably be to check the `mathMode` property of parents in the `sourceChildren` function.


PatrickMassot:
Thank you for your contribution. I am reluctant to merge such a large change without any new test files. 

It also seem to include two completely unrelated change sets. How is the mathtools change related to the description of the PR? Maybe this is a git mistake and you didn't intended it?

kno10:
The mathtools change adds some additional matrix types there, but you are right, it also includes a hack that may need some discussion. It is related though, and I did not have the time to untangle them. For intertext and shortintertext it tries to drop out of the alignment and recreate a new alignment afterwards, because there is no suport for intertext in MathJax (and IIRC this was one of the cases that broke the old logic even more). Feel free to skip these hunks.
For most of the change, the test given above should be useful to check for reduced newlines (double newlines are interpreted as paragraphs in markdown, so they are an issue).

PR #337, opened 2023-03-29
rudimentary support for \begin{enumerate}[start=4] (enumitem package)

eudoxos:
Please check this out, a sample document would be 
```tex
\documentclass{article}
\usepackage{enumitem}
\begin{document}
	\begin{enumerate}[start=4]
		\item foo 
		\item bar
	\end{enumerate}
\end{document}
```

which renders (HTML5 only) correctly as 

> 4. foo
> 5. bar

If you give it a green light, I would add unit tests, and support for some more enumitem options.

### DIFF
--- a/plasTeX/Packages/enumitem.py
+++ b/plasTeX/Packages/enumitem.py
@@ -1,0 +1,14 @@
+from plasTeX.Base.LaTeX.Lists import List
+class enumerate_(List):
+    macroName = 'enumerate'
+    args = '[ options:dict ]'
+    def invoke(self, tex):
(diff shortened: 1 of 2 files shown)


eudoxos:
Where I need help: with `\begin{enumitem}[label=\Roman*.,...]`, I need to get the `label` value before expansion, so that I can check whether it contains `\Roman`, `\alph` etc. Is it possible?

eudoxos:
(The CI failure originates in this master commit: https://github.com/plastex/plastex/actions/runs/4247580725)

PR #346, opened 2023-10-14
Fix image cache

kno10:
Written as binary, read as system encoding. use "rb" and "wb" consistently.

https://github.com/plastex/plastex/blob/408b07a1f8cdf47df76056141498d704510add6a/plasTeX/Imagers/__init__.py#L683


PatrickMassot:
Thanks!

PR #349, opened 2023-12-13
HTML5: renderer: Changes 'equation' content to mathjax_source

ianjauslin-rutgers:
In HTML5 renderer, 'equation' environment: Changed: 'obj.source' to 'obj.mathjax_source'.

I ran into an issue related to this when using inequalities ('<' or '>') in equation environments. It seems that using 'obj.source', these do not get escaped, and lead to problems. This patch switches 'obj.source' to 'obj.mathjax_source', and thus makes plastex escape '<' and '>' in 'equation' environments.

### DIFF
--- a/plasTeX/Renderers/HTML5/Math.jinja2s
+++ b/plasTeX/Renderers/HTML5/Math.jinja2s
@@ -13,7 +13,7 @@ name: equation
 <div class="equation" id="{{ obj.id }}">
 <p>
   <div class="equation_content">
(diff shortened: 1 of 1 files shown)


PatrickMassot:
Thanks. This was clearly an oversight when `.mathjax_source` was introduced. All other templates around use that.

ianjauslin-rutgers:
That's what I figured. Thank you!

PR #356, opened 2024-02-01
imath and jmath are MathSymbols

christianp:
\imath and \jmath were defined as MathAccents, which meant they required an argument. In fact, they're just symbols and don't take an argument.


PatrickMassot:
Good catch! git blame confirms this mistake has been around for at least 20 years... 

Thanks a lot for having taken time to write tests. This makes my work a lot easier.

PR #357, opened 2024-02-07
amsthm: change \newtheorem to take any TeXFragment as a header

christianp:
I encountered a LaTeX file where the header arguments to `\newtheorem` were wrapped in an extra pair of curly braces. pdflatex was fine with this, but plasTeX wasn't: it expected a plain string, so since the header argument was a `bgroup` element, it was lost.

This commit changes the type of the header argument to take any TeX fragment, and then takes its `textContent`.

### DIFF
--- a/plasTeX/Packages/amsthm.py
+++ b/plasTeX/Packages/amsthm.py
@@ -159,13 +159,13 @@ class newtheoremstyle(Command):
 
 
 class newtheorem(Command):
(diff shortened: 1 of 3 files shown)


PatrickMassot:
Ok, this is a weird edge case but handling it doesn’t seem to cost much.

christianp:
Thanks!

Issue #358, opened 2024-02-22
Dynamically find and load PlasTeX plugins


stephengaito:
PlasTeX should be able to dynamically find and subsequently load PlasTeX plugins.

At the moment PlasTeX provides the `--plugins` MultiString configuration option.

Unfortunately the Renderer options need to be loaded _before_ the command line options are parsed.

This is fine for existing Renderers which are distributed with the PlasTeX code, but the current setup does not allow PlasTeX plugins to independently distribute code whose Renderers provide their own (additional) command line options.

In particular the Gerby project has had to maintain their own fork of PlasTeX to get around these limitations (among other reasons).

It would be _preferred_ to enable the Gerby project to provide a focused PlasTeX plugin which only requires Gerby specific code changes. For example see Issue #160

stephengaito:
Potential work towards this goal can be found in the https://github.com/diSimplex/plastex fork of this project.

stephengaito:
At the moment ([4b66cd7](https://github.com/plastex/plastex/commit/4b66cd7442f9df8fe5eecb4deb4b7f9eb7ae8657)) the `config['general']['plugins']` list of plugins from the `--plugins` multi-string options are used in three places:

- Context.py (used in the `loadPythonPackage` method to load the `Packages` and run the associated `ProcessOptions`)
- Config.py (definition in the `defaultConfig` method -- _no change needed_ )
- Compile.py (used in the `load_renderer` method to dynamically load and return Renderer )

In addition `collect_renderer_config` method in the `client.py` file _should_ be able to dynamically find PlasTeX plugins and load their Renderer options.

---

To ensure minimal changes, these four methods will be updated to allow dynamically found PlaxTeX packages to be used _as well_ .

(No existing functionality, while potentially redundant, will be removed).

---
To implement these changes we will follow the method outlined in [Creating and discovering plugins - Python Packaging User Guide](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-naming-convention).

To do this we expect all PlasTeX plugins to start with `plastex_` and end with `_plugin`.

PR #359, opened 2024-02-24
tox tested minimal changes to support plastex issue 358

stephengaito:
This PR adds the ability to discover and load any Python package previously installed which follows a (new) PlasTeX plugin naming policy: `plastex_XXXX_plugin`. 

This PR provides a potential solution to the PlasTeX issue plastex/plastex#358.

It makes a very small **change in the Config.py file**, to add a new `--add-plugins` option which currently defaults to False, hence providing almost current behaviour. 

It also makes rather more **changes to the client.py file**. These changes do two things:

1. **Before command line options are parsed:** Discover any python packages which start with `plastex_` and end with `_plugin`. *If* the found package contains the python file `Renderers/Config.py` which itself contains an `addConfig` method, *then* this `addConfig` method is run with the PlasTeX `config` structure as its only argument, hence allowing the PlasTeX plugin to add command line options.
  - This functionality is currently *not* provided by the existing "plugins" system, and is needed to help, for example, split the Gerby specific code into a plugin which can (eventually) work with the existing (and heavily updated) PlasTeX code base.
2. **After command line options have been parsed:** Implements the `--add-plugins` option (if found) by *adding* the names of all installed PlasTeX plugins to the end of the existing PlasTeX `--plugins` option (list of strings). Hence all of the existing plugins implementation is essentially unchanged.

These changes ensure that existing behaviour is kept (except for the addition of a plugin's Renderers Options to the PlasTeX command line arguments).

It leverages the existing `plugins` code as it current exists to load TeX styles and/or Renderers.

---

**Testing using existing PlasTeX tests:** 

This PR has been tested (as the `plastex_issue_358` branch) using Tox (as requested in [plastex CONTRIBUTING document](https://github.com/plastex/plastex/blob/master/CONTRIBUTING.md)) using the [diSimplex/plastex-tox](https://github.com/diSimplex/plastex-tox) tool. 

(This `plastex-tox` tool tests a version of PlasTeX inside a specified Podman (Dockerfile) instance with known installation, using `tox` as described by the [plastex/plastex/tox.ini](https://github.com/plastex/plastex/blob/master/tox.ini) file).  

In all cases (python 3.7-3.12) there were two failures in *both* the existing PlasTeX repo and the `plastex_issue_358` branch of this repo. (Python 3.6 can no longer be tested out of the box with the new Tox versions).

In all failures there seems to be a problem with the `tikz`/`lualatex` and `pdf2svg` provided commands in the `plastex-tox` Dockerfile description (which was modelled upon the most recent Github action tests for PlasTeX).

Since this PR has no interaction with either `tikz`/`lualatex` or `pdf2svg`, I conclude that there is a problem with the testing environment (any pointers to correct this would be welcome) as opposed to any problems introduced by this PR.

### DIFF
--- a/plasTeX/Config.py
+++ b/plasTeX/Config.py
@@ -64,6 +64,12 @@ def defaultConfig(loadConfigFiles: bool=False):
         default = [],
     )
 
+    general['add-plugins'] = BooleanOption(
+        """Add all installed PlasTeX plugins.""",
+        options = '--add-plugins',
+        default = False,
+    )
+
     general['plugins'] = MultiStringOption(
         """
         A list of plugins to use. Each element must be a valid python module name,
--- a/plasTeX/client.py
+++ b/plasTeX/client.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python3
 
 import os, sys
-import importlib
+import traceback
+import importlib, pkgutil
 import traceback, pdb
 import plasTeX
 from plasTeX import __version__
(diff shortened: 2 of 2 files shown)


stephengaito:
Unfortunately, while this patch does use one of the three approaches to finding plugins as suggested in the [Creating and discovering plugins - Python Packaging User Guide](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-package-metadata), I have found that the use of `pkgutil.iter_modules` to be *very* unstable.

Reading the underlying Python code, the `pkgutil.iter_modules` *imports* *every* module found *as code*. This leads to very unstable (and seemingly random) behaviour. This approach is not appropriate for PlasTeX's use.

As such I am closing this PR with the intention of opening a newer PR which uses the (third and last) "Using package metadata" approach suggested in [Creating and discovering plugins - Python Packaging User Guide](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-package-metadata). 

This "meatadata" approach uses `importlib.metadata.entry_points`, which (again reading the underlying Python code) uses the pathlib.Path and zipfile.Path tools to scan the filesystems (and zip-filesystems) *as files*.  This "metadata" approach only loads each installed package's "metadata" files *as files*. This should be a much more "stable" approach to finding PlasTeX plugins.

PR #367, opened 2024-04-02
The enumerate environment ignores children which are not \item

christianp:
There was a very old case in the List environment's digest method which would ignore child nodes whose name is not 'item'. It looks like this was added so that unimplemented macros at the start of an `enumerate` environment don't produce list items in the output.

The List class is now used by quite a few other macros, which do allow other types of children. In particular, the bibliography unit tests failed when I tried just uncommenting the code.

This adds an `allowedChildren` attribute to List, which is `None` by default. When it's an iterable of node names, only nodes with those names are allowed, and all others are ignored. The `enumerate` class sets `allowedChildren = ('item',)`.

### DIFF
--- a/plasTeX/Base/LaTeX/Lists.py
+++ b/plasTeX/Base/LaTeX/Lists.py
@@ -16,6 +16,7 @@ class List(Environment):
     depth = 0
     counters = ['enumi','enumii','enumiii','enumiv']
     blockType = True
+    allowedChildren = None # A list of names of nodes that can be direct children of this node.
 
     class item(Command):
         args = '[ term ]'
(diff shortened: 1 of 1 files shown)


christianp:
Whoops, I wrote a unit test but didn't check it in. Will do!

PatrickMassot:
I don’t understand the test. What you wrote is not accepted by pdflatex, it triggers the infamous `! LaTeX Error: Something's wrong--perhaps a missing \item.`. Why do you expect plasTeX to handle that?

christianp:
Very good question! I can't find the real-world example that brought this problem to light.

The best I can find is an enumerate environment that starts `\begin{enumerate}\setlength{\itemindent}{3pt}`, but plasTeX seems to deal with that already. 

I'll just close this, until I work out what the problem actually was.

PR #368, opened 2024-04-02
Add support for custom item labels to enumerate

christianp:
(This builds on the changes added in #367)

This adds support for custom item labels to enumerate environments, following the format used by the enumerate LaTeX package.
    
The enumerate environment takes a single option argument specifying the format for item labels. The format string can contain any characters; the special characters I, i, A, a and 1 are replaced with corresponding encodings of the item's position in the list.

We tried a few different ways of rendering custom labels in HTML:

* `symbols()` in CSS: Chrome doesn't support this for anonymous list counters
* `@counter-style` rules: produced very complicated HTML for standard lists, and there are some compatibility issues

A decision has to be made about whether custom labels should be selectable: default list labels in HTML aren't selectable, but you could make a case that custom labels are part of the text and should be selectable. Certainly, we saw some cases where authors used custom labels for text that you'd expect to be included when you copy the list to the clipboard.

Finally, we just made separate container elements for the item's label and content, and used CSS subgrid display to lay them out. `display: subgrid` is very new, so you might consider this too adventurous. I haven't included our CSS in this branch, so this shouldn't be merged until there's agreement on how these should be displayed.

The HTML5 templates now also include `id` attributes and set the `value` attribute for list items, to make them linkable and to help assistive tech. The benchmark unit tests for a few things that use `enumerate` have had to be updated, as a consequence.

### DIFF
--- a/plasTeX/Base/LaTeX/Lists.py
+++ b/plasTeX/Base/LaTeX/Lists.py
@@ -5,6 +5,7 @@ C.6.3 The list and trivlist Enviroments
 """
 
 from plasTeX import Environment, Command, DimenCommand
+from plasTeX.encoding import numToRoman, numToAlpha
 
 class enuminame(Command): str = ''
 class enumiiname(Command): str = ''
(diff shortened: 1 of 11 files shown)


christianp:
Sorry, the enumerate unit test runs on its own but not when run with everything else. I'll try and work out why.

christianp:
@PatrickMassot This branch was based on top of the `enumerate-ignore-non-items` branch, but didn't need to be. I've rebased it on top of master and it looks OK.

PR #371, opened 2024-04-02
implement \rightline

christianp:
The \rightline command causes its contents to be aligned to the right.

For HTML5 output, this is equivalent to \raggedleft or \flushright.


PatrickMassot:
Thanks!

PR #373, opened 2024-04-03
pdf2svg imager removes beamer background color and nav

christianp:
If the document uses the beamer documentclass and sets a background color or displays navigation elements, these would be included in the SVG imager's output, preventing the image from being cropped correctly.

This commit adds a bit to the preamble for the pdf2svg imager to clear the beamer background.

This also adds a line to the imager unit test and the new beamer imager unit test to add the unittests directory to sys.path, so they can be run on their own. Previously, they'd throw an error "no module helpers" if run as single files.


PatrickMassot:
Thanks. I wasn’t too happy with the generic imager code to contain beamer-specific code so I implemented a more specific solution in 65481133239a18fcc226568d655574ef58a7c28a.

christianp:
That's much better, thanks!

PR #374, opened 2024-04-03
add a dummy amsbsy package

christianp:
The amsbsy package provides math-mode commands \pmb and \boldsymbol. In HTML, MathJax implements these commands.

This dummy package prevents plasTeX from trying to load the real LaTeX amsbsy package.


PatrickMassot:
Thanks. Could you please change your “handled by mathjax” to a more generic “handled by math renderer”? Technically there is nothing ensuring MathJax will be used. The HTML5 renderer is not the only renderer.

christianp:
@PatrickMassot done, and I've added a CTAN link.

PR #375, opened 2024-04-03
add a dummy amscd package

christianp:
amscd draws commutative diagrams. It's implemented by MathJax, so this just adds a dummy package to prevent plasTeX from trying to load the real LaTeX package.


christianp:
I haven't written a unit test for this, because I'm not sure what to test: should I just try processing a document that uses `amscd`, and check that it renders at all?

PR #376, opened 2024-04-03
add the bbm package

christianp:
The bbm package provides a "blackboard bold" font, and a `\mathbbm` command to use that font.

MathJax supports these characters in the `\mathbb` command, so `\mathbbm` is rewritten to `\mathbb`, to be handled by MathJax.


PatrickMassot:
Thanks! Could you please add a short documentation including a link to ctan and a sentence explaining this will be turned into a `mathbb`?

christianp:
@PatrickMassot done!

PR #377, opened 2024-04-03
add the cancel package

christianp:
MathJax implements the cancel commands, so plasTeX just needs to pass them through, and not load the LaTeX version.

### DIFF
--- a/plasTeX/Packages/cancel.py
+++ b/plasTeX/Packages/cancel.py
@@ -1,0 +1,14 @@
+from plasTeX import Command
+
+
+class cancel(Command):
+    args = 'self'
+
+class bcancel(Command):
+    args = 'self'
+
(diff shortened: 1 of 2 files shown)


PatrickMassot:
Thanks! 

The commands `\cancel`, `\bcancel` and `\xcancel` actually also work in (horizontal) text mode. One cheap way to emulate this would be to create html templates using the strikethrough html tag (for all three version). But this is probably a very uncommon case so I would understand if you don’t want to invest time doing this.

In both cases, it would be nice to add documentation to the new package file, including a link to ctan and a comment if you don’t add support for the text mode case.

christianp:
Good suggestion, @PatrickMassot! I've done that, so `\cancel` in text is rendered as a `<del>` element.

PR #378, opened 2024-04-03
add support for \caption*, and dummy caption package

christianp:
The longtable and caption packages add a command `\caption*`, which produces a caption with no title or reference number.

This adds an empty file caption.py, to prevent plasTeX loading the LaTeX caption package, and changes the definition of the Caption command to allow a starred version.


PatrickMassot:
Thanks! Could you please write your PR description explanations in the new dummy file?

PR #385, opened 2024-07-14
Fix mathjax-macros

eric-wieser:
This was always a javascript syntax error due to a missing comma between dictionary entries.
Trailing commas are fine.


eric-wieser:
@PatrickMassot, thanks for triggering the unit tests; hopefully they should now all pass this time.

PR #388, opened 2024-09-13
cope with # as a delimiter for \verb

christianp:
The character # is parsed as a `plasTeX.Tokenizer.Parameter` object when it immediately follows `\verb`, but is parsed as a `plasTeX.Tokenizer.Other` object otherwise.

This breaks the logic for the `\verb` command, which checks tokens to see if they're equal to the token given as the delimiter.

This commit adds a special case for `Parameter` objects as the end pattern, and changes it to an `Other` object.

There's a unit test verifying that `\verb#...#` works as it should.


PatrickMassot:
Could you please add documentation in the code you modified? I think it will be pretty hard to understand after forgetting about this PR.

christianp:
@PatrickMassot good suggestion, thanks. I've added a few lines of comments; hopefully that's enough to explain what's going on.

PR #392, opened 2024-09-29
Enable testing with Python 3.13

llimeht:
This PR

- enables testing with Python 3.13 (rc)
- updates the versions of the actions to reduce noise from CI
- lets the python action cache pip downloads more aggressively
- fixes compatibility with Python 3.13

The original purpose of this work was to address the removal of the `cgi` module in Python 3.13 - that fix is in the final commit, where `html.escape` replaces `cgi.escape`. However... while the `cgi` module only gets removed in Python 3.13, the actual `cgi.escape` function was removed in Python 3.8. 

There's not much test coverage of `simpletal` for this to have gone unnoticed for so long (see below). Presumably there are also no users of `simpletal` given no-one has complained in 5 years. This PR is safe to merge as is, but perhaps it's appropriate to just remove `simpletal` in a future release.

Test coverage:
```
plasTeX/Renderers/PageTemplate/simpletal/FixedHTMLParser.py                                                                                4      1    75%
plasTeX/Renderers/PageTemplate/simpletal/__init__.py                                                                                       1      0   100%
plasTeX/Renderers/PageTemplate/simpletal/sgmlentitynames.py                                                                                1      0   100%
plasTeX/Renderers/PageTemplate/simpletal/simpleTAL.py                                                                                    941    800    15%
plasTeX/Renderers/PageTemplate/simpletal/simpleTALES.py                                                                                  436    367    16%
plasTeX/Renderers/PageTemplate/simpletal/simpleTALUtils.py                                                                               128    128     0%
```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Use Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           allow-prereleases: true
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,6 +19,8 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           allow-prereleases: true
+          cache: pip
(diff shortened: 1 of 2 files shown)


eric-wieser (approved):

PatrickMassot:
Thanks. simpletal is indeed completely obsolete and I should remove it in the next release.

PR #398, opened 2025-01-31
fix the \ding command

christianp:
`\ding` seems to come from the pifont package, but has been implemented in plasTeX in the Base.LaTeX.Characters package.

It should produce a dingbat character, which is encoded in the Dingbats block in Unicode - see https://en.wikipedia.org/wiki/Dingbats_(Unicode_block)

There was a bug which led to an infinite recursion, since the \ding command implements the `str` property as a method which refers to `self.textContent`, which begins by looking at `self.str`.

This commit adds a different implementation of `textContent` which doesn't look at `self.str`, but I'm not sure if this would ever be used if the command is used as designed, since this commit also adds a mapping from Zapf dingbats to Unicode.


PatrickMassot:
Thanks!

PR #399, opened 2025-02-20
Fix bugs in kpsewhich

eric-wieser:
When `kpsewhich` was not installed, this could:
* not resolve absolute paths.
* not resolve paths of the form `foo/bar.tex`

Additionally, this would sometimes swallow `KeyboardInterrupt`, and would sometimes not clean up `TEXINPUTS` properly.
The cleanup is now handled with `contextlib.ExitStack`, which simplifies the return paths and ensures that the cleanup is run even if the code is interrupted with KeyboardInterrupt.

### DIFF
--- a/plasTeX/TeX.py
+++ b/plasTeX/TeX.py
@@ -1357,16 +1357,16 @@ class TeX(object):
 
             try:
                 program = self.ownerDocument.config['general']['kpsewhich']
-    
+
                 kwargs = {'stdout':subprocess.PIPE}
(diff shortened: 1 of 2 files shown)


PatrickMassot:
Thanks Eric. Could you please add tests?

eric-wieser:
@PatrickMassot, please let me know if you'd prefer me not to ping you on plastex-related things in future; but the tests are there and pass.

PR #403, opened 2025-04-02
strip trailing whitespace from the command name in \[re]newcommand

christianp:
By default, plasTeX adds a space character to the end of the source of any element that doesn't have any arguments, to ensure that it's separated from the next token.

Most of the time this is safe because LaTeX usually ignores extra whitespace. But the `\renewcommand` command is sensitive to extra whitespace in the name of command being defined, when it's not a letter character, e.g. `\renewcommand{\> }` is not interpreted the same as `\renewcommand{\>}`.

I think that all of the `.source` methods on elements assume that they don't need to know their outer context, so it was hard to think of a good way of fixing this.

This commit adds a method `filterArgumentSource` to `Macro`, which is called on the source of each argument as it's processed.

The `\newcommand` macro and all the others that inherit from it use this to remove any whitespace from the inside of the first argument.

### DIFF
--- a/plasTeX/Base/LaTeX/Definitions.py
+++ b/plasTeX/Base/LaTeX/Definitions.py
@@ -23,6 +23,12 @@ class newcommand(Command):
         deflog.debug('command %s %s %s', *args)
         self.ownerDocument.context.newcommand(*args, **kwargs)
 
+    def filterArgumentSource(self, source, i):
+        if i==1:
+            source = source.replace(' ','')
+
+        return super().filterArgumentSource(source, i)
+
 class renewcommand(newcommand):
     pass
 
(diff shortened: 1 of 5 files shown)


christianp:
I think the motivation for this was an author who had redefined the macros `\<` and `\>` to do something non-standard, and plasTeX was failing to pass the definitions through to MathJax because it was adding space. That's hard to write a unit test for because the difference is only apparent in a browser, once MathJax runs.

christianp:
I've found what the original problem was: it wasn't anything to do with MathJax, but instead with the imager mechanism, which we can unit test. 

I've added a new unit test, `imagers/renewcommand.py`, which demonstrates this.

PR #404, opened 2025-04-17
try to make includegraphics insensitive to the case of the file's extension

christianp:
A package had a file `graphic.JPG`. The command
`\includegraphics{graphic}` didn't find it, because it was only looking for `graphic.jpg`.

This changes the `\includegraphics` command to glob all files whose names start with the given name, and then return the first file whose extension, converted to lower case, is in the expected list.


PatrickMassot:
I’m not sure about this change. I think your new code will always fail if the provided path includes a directory, say `\includegraphics{myfolder/myimage}`. Also if you add `''` in front of `ext` then it becomes unnecessary to add it again a few lines later.

